PR #2, opened 2025-08-22
Use CHERIoT-Platform's LLVM + CI to check it

xdoardo:
I think I disabled the CI tasks from rust-lang/rust. I want to check if it is true.
I might use this PR to do some tests: I'd like to see if we can have a simple CI job that does sanity checks, such as  running `tidy` and compiling `rustc` on a selection of target machines using CHERIoT-Platform's LLVM. 

Edit: this is done by changing the default remote and branch of the llvm-project submodule in Rust to the ones from CHERIoT-platform, which might be a debatable choice. Also, notice that this commit only sets up the submodule and adds sanity checks in CI, but does not change the Rust/C++ code in rustc that re-exports and links to the LLVM API: CI is therefore bound to fail for now :)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,59 @@
+gcp_credentials: ENCRYPTED[5ee9e22237c8368fdae53b368a38509d62a168e1b5a8162a9bd8dfef5e874a4f41cd8235d498b99d2c2860c77e06ecd1]
+
+# -- PR tasks
+
+pr_check_commits_task:
+  skip: $CIRRUS_PR == ""
+  name: (PR) Check commits structure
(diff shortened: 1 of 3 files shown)


xdoardo:
Alright. CI now has two jobs: the first one checks that the PR adds a single commit; this is useful to keep track of updates to `beta`, to make cherry-picking to `master` easier. We might also add a check on the message of the commit (e.g. starts with "cheri:" to distinguish what's ours from what comes from upstream). The second job checks out the source code and runs `./x check`.

seharris (approved):
Looks ok to me.
Do we need to talk about what CI service to use at some point?

PR #15, opened 2025-09-02
Add the `riscv32cheriot-unknown-cheriotrtos` target

xdoardo:
This PR introduces the new `riscv32cheriot-unknown-cheriotrtos` target. Of course, this PR doesn't add _full support_ for the target just yet: some parts - most parts - are expected to break when targeting this platform. 

#### A note about the name
As per [this](https://discourse.llvm.org/t/rfc-support-cheriot-target-triple-as-subarchitecture-of-riscv32/87883/10) discussion, it seems that the nomenclature `riscv32cheriot-unknown-cheriotrtos` - currently supported by LLVM - will be unsupported in the future. `clang`, for example, will need more flags (`-mcpu`, `-march`, potentially others?) to correctly identify the same target.  

Rust, as far as I can tell, has instead a simpler way to refer to targets:  `rustc` only has the `--target` parameter, not `-mabi` and the likes. Therefore, while [internally](https://github.com/CHERIoT-Platform/cheri-rust/compare/beta...xdoardo:add-cheriot-target?expand=1#diff-809e05976a191a4195a6d4a755080b90eb6c141896a553cb5f7387f200c6c2e2R10-R11) we indeed use the "correct" nomenclature, it seems to me that it makes more sense that in _Rust land_ we use a nomenclature to precisely identify this target. 

`riscv32-unknown-cheriotrtos` could be an alternative, but it seems to me that it is less explanatory than it should, in this context.


xdoardo:
I was wrong: `rustc` has the `-C target-cpu=xyz` and `-C target-feature=+x,-y` [flags](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-cpu).

For now, the choice is just stylistic, and `riscv32cheriot-unknown-cheriotrtos` doesn't entail a lot of changes v.s. `riscv32-unknown-cheriotrtos` since, of course, CHERIoT is the only (or the default) architecture supported by `cheriotrtos`. Therefore, changing the name to remove CHERIoT wouldn't entail changes to the default definition of the target.

It would be different for baremetal CHERIoT, which in LLVM has the triple `riscv32-unknown-unknown`. In `clang`, this target is identified through the use of the appropriate flags (I would imagine something like `-march=cheriot`?); in `rustc`, this could use the `riscv32e-unknown-none-elf` base target, and users would have to at least specify `-C target-cpu="cheriot"`. 

This would also mean that, in `rustc`, we have to check whether the user-given `target-cpu` and/or `target-feature` flags imply that the target platform is CHERIoT, and adapt the relevant `Target` fields (data layout string, default address space, `families`.. ) accordingly: this, to me, sounds a bit messy. 

We can postpone the decision until we actually get to the implementation of baremetal CHERIoT, but it could be useful to have an idea of what would happen then. @seharris, do you happen to already have an opinion on this?

seharris (changes requested):
I think there's a minor error in the documentation (see my comment). Apart from that, this looks good.

PR #32, opened 2025-10-27
Create a runner to execute tests in the `cheri/tests` directory

xdoardo:
This PR adds a test runner specific to the CHERIoT platform. 

This runner depends on a CHERIoT-enabled `rustc`, `xmake`, `git` and a CHERIoT-enabled LLVM. It takes all the `.rs` files in the `cheri/tests/**/*.rs` directory (blocking the subpaths starting with the runner itself) and links each of them in a CHERIoT firmware (one firmware for each test) using `xmake` and the CHERIoT-RTOS components. It then runs each firmware image using `xmake run`. 

(In progress: CI)


cheri/tests/runner/libcheriot/README.md
```diff
@@ -0,0 +1,6 @@
+# A bare-bones support library for running Rust tests on CHERIoT
+
+You need to define:
```
seharris: I don't really understand what's going on here, is something missing?

xdoardo: Oops, it's a leftover from [this](https://github.com/xdoardo/libtest-barebones) other `libtest` clone for CHERIoT.

seharris (approved):

seharris:
One minor comment, there are a couple of places the CI script clones Git repos (like the Sail model) where I think we could use `--depth=1` to speed up clones and reduce bandwidth requirements a bit. In theory `--shallow-submodules` might be useful too, but it sounds like that can cause trouble for submodules with un-tagged commits.

PR #34, opened 2025-11-07
Add `is_like_cheri` and use it to translate inline asm of `black_box`

xdoardo:
The `is_like_cheri` target option must be used if and only if the families of the target contain `"cheri"`.

This value is then used to make the compiler generate a slightly different inline assembly when translating the `black_box` intrinsic which, as it was before, would cause an error on CHERIoT:
```
error: couldn't allocate input reg for constraint 'r'
```
With this patch, the generated inline assembly uses as constraint `"C"` (capability-sized register) instead of `"r"`. This patch should solve this problem not only on CHERIoT, but hopefully on other CHERI-like platforms as well: we can't, however, test this claim right now, as we currently support CHERIoT only.

Also adds a test to verify that basic usage of `black_box` works as expected.

### DIFF
--- a/compiler/rustc_codegen_llvm/src/intrinsic.rs
+++ b/compiler/rustc_codegen_llvm/src/intrinsic.rs
@@ -18,6 +18,7 @@ use rustc_middle::{bug, span_bug};
 use rustc_span::{Span, Symbol, sym};
 use rustc_symbol_mangling::{mangle_internal_symbol, symbol_name_for_instance_in_crate};
 use rustc_target::callconv::PassMode;
+use rustc_target::spec::HasTargetSpec;
 use tracing::debug;
 
 use crate::abi::FnAbiLlvmExt;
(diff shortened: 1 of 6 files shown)


compiler/rustc_codegen_llvm/src/intrinsic.rs
```diff
@@ -515,6 +516,8 @@ impl<'ll, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'_, 'll, 'tcx> {
                 // the memory.
                 let (constraint, inputs): (&str, &[_]) = if result.layout.is_zst() {
                     ("~{memory}", &[])
+                } else if self.target_spec().is_like_cheri {
+                    ("C,~{memory}", &result_val_span)
```
seharris: Would it be worth adding a comment explaining that (if I remember the conversation correctly) we think this value is always a pointer, hence use of the C register class?

xdoardo: I added a comment, let me know if you think it's clear enough.

seharris (approved):

PR #41, opened 2025-11-24
Use custom readme and add example

xdoardo:
To make it a little bit easier for people to approach this project. 

The complexity of the infrastructure of the example shows some aspects where we need to get better, I think.


jacobpake:
Nice! I will have a go at building the example on a fresh box tomorrow and keep a note of the steps.

README.md
```diff
@@ -1,77 +1,85 @@
...
+ln -s $PWD/build/host/stage1-tools-bin/cargo $RUSTUP_HOME/toolchains/cheri/bin &&\
+ln -s $PWD/build/host/stage1-tools-bin/rust-analyzer $RUSTUP_HOME/toolchains/cheri/bin
+```
+
+### Now what?
+
+Keep in mind that this is an experimental project, and anything can break at any time.
+
+As of now, the process of getting a final image for CHERIoT will entail
+1. Having in your Rust project unmangled `extern "C"` API;
```
seharris: I found this wording a bit hard to parse, I think I'd have said something like, “Providing an umangled `extern "C"`  API in your Rust project”.

jacobpake: > Fixed in [09cf648](https://github.com/CHERIoT-Platform/cheri-rust/pull/41/commits/09cf648dbf989b29b16f794b75d3387de56092ac)

It would be useful to see the individual commit history during PR review - can I suggest keeping commits in the history, and squashing only when we are ready to merge? Not sure if there is a standard way to handle this

xdoardo: I think it makes sense for larger changes. It personally works fine for me for small changes like this one where clicking on the comment takes you to the updated line. I am not opposed to it though, I'll start doing that if it makes it easier.

seharris: Yea, the link to [09cf648](https://github.com/CHERIoT-Platform/cheri-rust/pull/41/commits/09cf648dbf989b29b16f794b75d3387de56092ac) is also 404 for me.

Other than that though, if @jacobpake didn't have any problems running the example, I think this should be ready to merge.

jacobpake (approved):

PR #45, opened 2025-11-28
Pack size and alignment in vtables for CHERIoT

### LINKED ISSUE #48: ICE when generating `vtable` debug info for CHERIoT
The following assertion
https://github.com/CHERIoT-Platform/cheri-rust/blob/fb4b1bc22072a122a03e658231a1d719458fc68e/compiler/rustc_codegen_llvm/src/debuginfo/metadata.rs#L1463
fails when targeting CHERIoT. 
 
### Description
Rust generates virtual tables making the assumption that every entry in it will have the same in-memory size, in particular that of a pointer. This makes sense for all the platforms with integral pointers, but makes slightly less sense for targets like CHERIoT, because the size and the alignment - two `usize`s -  of the type the virtual table is tied to are stored as entries in the table as well. The presence of those values are, as far as I can tell, exactly the reason why that assertion is there.

Note that this also means that the tables are laid out "wasting" 64 bits of space in total when targeting CHERIoT, because the size and alignment are automatically padded to 64 bits, while they really are 32 bits. Take, for example, the following trait and implementation:
```rust
trait Chair {
    fn count_legs(&self) -> u32;
}

struct Monobloc {
    dropme: String
}

impl Chair for Monobloc {
    fn count_legs(&self) -> u32 {
        4
    }
}
```
The following vtable is generated for `Monobloc`
```llvm
@vtable.0 = private unnamed_addr addrspace(200) constant <{ ptr addrspace(200), [4 x i8], [4 x i8], [4 x i8], [4 x i8], ptr addrspace(200) }> <{ ptr addrspace(200) @"_ZN4core3ptr37drop_in_place$LT$vtable..Monobloc$GT$17hd4707355d5a501fdE", [4 x i8] c"\18\00\00\00", [4 x i8] undef, [4 x i8] c"\08\00\00\00", [4 x i8] undef, ptr addrspace(200) @"_ZN50_$LT$vtable..Monobloc$u20$as$u20$vtable..Chair$GT$10count_legs17h0d687f8285288760E" }>, align 8
														------------------------| size |  padding  | align |  padding  |
                                                                                                                                               -------------------------------------------------------------------------------------------------|          size         |    padding     |           align         |    padding    |
```

#### Potential solutions

I think we can either turn off that assertion, or keep it there but kind of cheat and replace the `pointer_size` to refer to the capacity of the pointer (i.e. the size of an address), or remove the assertion and try to make rust generate the virtual tables correctly for CHERIoT as well (i.e. packing size and alignment in 64 bits). I think the second solution is worth exploring, but depending on how much change it entails it might be hard to upstream.

xdoardo:
A potential solution for #48 - still a draft because I want to add tests to the generated LLVM IR for CHERIoT before merging, and I would also like to have #59 fixed before merging this one.

The patch adds `in_memory_size` and `data_size` methods to the `VtblEntry` type so to make computations respect the different sizes of the entries on CHERI-like platforms (or any platform with non-integral pointers). These methods are used both to generate the vtable type in LLVM IR and to generate the debug info for the type.

For reference, this is what the vtable in the example in the issue now looks like:
```llvm
@vtable.0 = private unnamed_addr addrspace(200) constant <{ ptr addrspace(200), [8 x i8], ptr addrspace(200) }> <{ ptr addrspace(200) @"_ZN4core3ptr37drop_in_place$LT$vtable..Monobloc$GT$17h68c892b65ff866f1E", [8 x i8] c"\18\00\00\00\08\00\00\00", ptr addrspace(200) @"_ZN50_$LT$vtable..Monobloc$u20$as$u20$vtable..Chair$GT$10count_legs17h0a91b89070da3234E" }>, align 8
```

#### Some questions
* How much will this patch degrade performance on targets where these computations are just trivial multiplications? Should we split the computation in two functions, a "trivial" one for usual platforms and an "expensive" one for platforms with non-integral pointers?
* Is there a way to make sure that we're adding reasonable checks for other platforms with non-integral pointers, such as making sure the accesses to the vtable are aligned?

### DIFF
--- a/compiler/rustc_codegen_ssa/src/meth.rs
+++ b/compiler/rustc_codegen_ssa/src/meth.rs
@@ -28,7 +28,19 @@ impl<'a, 'tcx> VirtualIndex {
 
         let llty = bx.fn_ptr_backend_type(fn_abi);
         let ptr_size = bx.data_layout().pointer_size();
-        let vtable_byte_offset = self.0 * ptr_size.bytes();
+        let usize_size = bx.data_layout().pointer_offset();
+
+        // FIXME(xdoardo) Find a way to remove the overhead for this comparison too?
+        let vtable_byte_offset = if ptr_size == usize_size {
+            ptr_size.bytes() * self.0
+        } else {
+            match self.0 {
+                0 => 0,
+                1 => usize_size.bytes(),
+                2 => usize_size.bytes() * 2,
+                n => usize_size.bytes() * 2 + ptr_size.bytes() * (n - 2),
+            }
+        };
 
         load_vtable(bx, llvtable, llty, vtable_byte_offset, ty, nonnull)
     }
--- a/compiler/rustc_codegen_ssa/src/meth.rs
+++ b/compiler/rustc_codegen_ssa/src/meth.rs
@@ -34,11 +34,16 @@ impl<'a, 'tcx> VirtualIndex {
         let vtable_byte_offset = if ptr_size == usize_size {
             ptr_size.bytes() * self.0
         } else {
-            match self.0 {
-                0 => 0,
-                1 => usize_size.bytes(),
-                2 => usize_size.bytes() * 2,
-                n => usize_size.bytes() * 2 + ptr_size.bytes() * (n - 2),
+            let idx = self.0 as usize;
+            if idx < TyCtxt::COMMON_VTABLE_ENTRIES.len() {
+                TyCtxt::COMMON_VTABLE_ENTRIES[..idx]
+                    .iter()
+                    .fold(0, |acc, item| acc + item.memory_size(bx).bytes())
+            } else {
+                TyCtxt::COMMON_VTABLE_ENTRIES
+                    .iter()
+                    .fold(0, |acc, item| acc + item.memory_size(bx).bytes())
+                    + (ptr_size.bytes() * (self.0 - TyCtxt::COMMON_VTABLE_ENTRIES.len() as u64))
             }
         };
 
--- a/cheri/tests/vtable.rs
+++ b/cheri/tests/vtable.rs
@@ -1,0 +1,55 @@
+#![no_std]
+
+extern crate alloc;
+extern crate cheriot;
+
+use alloc::string::String;
+
+trait Chair {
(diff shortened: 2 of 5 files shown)


xdoardo:
Just noticed a bug: 
```llvm
@vtable.0 = private addrspace(200) constant <{ ptr addrspace(200), [4 x i8], [4 x i8], [4 x i8], [4 x i8], ptr addrspace(200), ptr addrspace(200), ptr addrspace(200) }> <{ ptr addrspace(200) @"..", [4 x i8] c"..", [4 x i8] undef, [4 x i8] c"..", [4 x i8] undef, ptr addrspace(200) @"..", ptr addrspace(200) @"..", ptr addrspace(200) @.. }>, align 8, !dbg !7
```
is still wrong, it counts again size and alignment as `[4 x i8], [4 x i8], [4 x i8], [4 x i8]`, that is 64 bits each. I suspect this comes from a different function than the one I patched, but I might be wrong. It could be due to padding, but I think that LLVM should be smart enough to pack those two values in 64 bits. Then again, maybe this optimisation is skipped when building in debug mode..

compiler/rustc_codegen_ssa/src/meth.rs
```diff
@@ -28,7 +28,19 @@ impl<'a, 'tcx> VirtualIndex {
...
+
+        // FIXME(xdoardo) Find a way to remove the overhead for this comparison too?
+        let vtable_byte_offset = if ptr_size == usize_size {
+            ptr_size.bytes() * self.0
+        } else {
+            match self.0 {
+                0 => 0,
+                1 => usize_size.bytes(),
+                2 => usize_size.bytes() * 2,
+                n => usize_size.bytes() * 2 + ptr_size.bytes() * (n - 2),
```
seharris: I'm not clear where the magic number two comes from here, I think this needs a comment that explains what's happening. If this calculation is accounting for the entries in `TyCtxt::COMMON_VTABLE_ENTRIES` then it might also be nice to tie it into those values somehow, so any future changes there won't cause mysterious weirdness.

xdoardo:
We still want to fix this, but I am closing this PR because it's stale now and it's just a burden to update when updating our Rust version.

PR #67, opened 2025-12-29
cheri/runner: Use the rust-rules branch of the RTOS.

### LINKED ISSUE #54: Integrate `cargo` and/or `rustc` in the RTOS's build system
The [CHERIoT-RTOS](https://github.com/CHERIoT-Platform/cheriot-rtos) uses [xmake](https://xmake.io/) to build firmware images. 
We want to be able to instruct `xmake` to add a rust dependency to a compartment and link the resulting object files to it. 

Currently, users are forced to do something that is very likely to produce wrong firmware images, that is, manually adding flags to the linker to add the object files generated from Rust: all the configurations should instead happen with `xmake`'s [native way](https://xmake.io/examples/other-languages/rust.html) to interact with rust codebases.

xdoardo:
Should fix #54. It's useful to edit this and https://github.com/CHERIoT-Platform/cheriot-rtos/pull/634 together so that we have at least a bit of confidence that the new rules work!

This PR edits the runner to generate correct `xmake.lua` configurations files to be able to run the tests using the new rust-rules from the RTOS. This PR also updates the examples to use the same updated xmake configuration.

Edit: adding another commit (thus making CI red) to make CI also build and run the examples.


cheri/tests/runner/src/runner/mod.rs
```diff
@@ -168,17 +196,35 @@ impl App {
...
+        self.xmake_strict(dir, &["config", &sdk_flag, &rc_flag]).map(|_| ())
     }
 
     fn xmake_build(&self, dir: &Path) -> anyhow::Result<()> {
         traceln!(self, "running `xmake` configuration command");
         self.xmake_strict(dir, &[]).map(|_| ())
     }
 
+    fn xmake_build_libcheriot(&self, dir: &Path) -> anyhow::Result<()> {
+        traceln!(self, "running `xmake` configuration command");
```
jacobpake: As other comment

Issue #73, opened 2026-01-13
Missing types for atomic operations in `atomic-operations.rs`


xdoardo:
https://github.com/CHERIoT-Platform/cheri-rust/blob/aedfdf14762fd2ba811bf78a8088127c83736b57/tests/codegen-llvm/atomic-operations.rs#L1

This test fails to compile due to:
```
error[E0599]: no method named `compare_exchange` found for reference `&AtomicI32` in the current scope
  --> tests/codegen-llvm/atomic-operations.rs:15:15
   |
15 |     let _ = a.compare_exchange(0, 10, Relaxed, Relaxed);
   |
```

This is due to this configuration value:
https://github.com/CHERIoT-Platform/cheri-rust/blob/aedfdf14762fd2ba811bf78a8088127c83736b57/compiler/rustc_target/src/spec/targets/riscv32cheriot_unknown_cheriotrtos.rs#L28

Issue #74, opened 2026-01-13
Missing `AtomicPtr` type


xdoardo:
https://github.com/CHERIoT-Platform/cheri-rust/blob/aedfdf14762fd2ba811bf78a8088127c83736b57/tests/codegen-llvm/atomicptr.rs#L11
fails due to
```
error[E0432]: unresolved import `core::sync::atomic::AtomicPtr`
```
This is due to this line:
https://github.com/CHERIoT-Platform/cheri-rust/blob/aedfdf14762fd2ba811bf78a8088127c83736b57/compiler/rustc_target/src/spec/targets/riscv32cheriot_unknown_cheriotrtos.rs#L27

The fix is to set it on 64 bits, which CHERIoT supports for atomic operations.

seharris:
Also affects `cheri-rust/tests/codegen-llvm/noalias-rwlockreadguard.rs`

PR #106, opened 2026-02-10
Set `max_atomic_width: 64`, `atomic_cas: true` in `riscv32cheriot` 

xdoardo:
Because CHERIoT does support atomic operations on 64-bit values, and also supports atomic compare and swap operations. See, for more details, [this header](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/include/stdatomic.h) in the RTOS.

Fixes https://github.com/CHERIoT-Platform/cheri-rust/issues/73 and https://github.com/CHERIoT-Platform/cheri-rust/issues/74, albeit the last issue also concerns another test which uses types from the stdlib, which we can't fix now.


jacobpake:
For further reference:

[Atomic portability](https://doc.rust-lang.org/nightly/std/sync/atomic/index.html#portability)
[Semi-related discussion upstream (for multicore)](https://github.com/rust-lang/rust/issues/117305)

seharris:
Right, a quick poke around in the RTOS is all it took to confirm I'm right about how atomics work.

PR #107, opened 2026-02-10
Add CHERI-specific e_flags when targeting CHERIoT

xdoardo:
As per title. It seems to me that adding a specific test for this bit requires some changes to the infrastructure (for example installing objdump/bingrep and then run a shell command that greps for `e_flags: 0x70009` in the output) which I think would outweigh this specific patch.


src/tools/tidy/src/extdeps.rs
jacobpake: I didn't see this before, will `x tidy` fail without this?

jacobpake (approved):

Issue #108, opened 2026-02-11
`alloc::format!` with 'alternate' flag causing `PermitExecuteViolation` error


jacobpake:
Use of the alternate flag (a `#` which tells the formatter to prefix hex with `0x`, binary with `0b`, etc.) causes a `PermitExecuteViolation` error in the simulator.

Example:

```rs
let x = 42;
assert_eq!(alloc::format!("{x:#b}"), "0b101010");
```

>Error handler: PermitExecuteViolation(0x11) error at 0x8000f3fa (v:0 0x8000a100-0x80015400 l:0xb300 o:0x0 p: G R-cgm- X- ---) (return address: 0x8000eb20 (v:1 0x8000a100-0x80015400 l:0xb300 o:0x5 p: G R-cgm- X- ---)), with capability register CS1(0x9): 0x800177b8 (v:1 0x800177b8-0x800177d0 l:0x18 o:0x0 p: - RWcgml -- ---)

The following is fine:

```
let x = 42;
assert_eq!(alloc::format!("{x:b}"), "101010");
```

This effects several doctests in `core/fmt`.

Issue #84, opened 2026-01-21
Instruction selection failure `frem`


jacobpake:
Compiling

```rs
fn f64_rem(a: f64, b: f64) -> f64 {
    a % b
}
```

Throws

```
rustc-LLVM ERROR: Cannot select: 0x139023e80: f64 = frem 0x13902d310, 0x13902dcb0
  0x13902d310: f64,ch = CopyFromReg 0x128622458, Register:f64 %0
  0x13902dcb0: f64,ch = CopyFromReg 0x128622458, Register:f64 %1
```

Possible solution is to have this lower to an fmod libcall (and make that available via the RTOS)

PR #111, opened 2026-02-13
Bump LLVM and add alternate `fmt` test

xdoardo:
As per title. Investigation on https://github.com/CHERIoT-Platform/cheri-rust/issues/108 brought to light a bug in LLVM, where callee-saved registers could be used as target of indirect tail calls. This bug was fixed with https://github.com/CHERIoT-Platform/llvm-project/pull/323, which this PR pulls in. 

This update also pulls https://github.com/CHERIoT-Platform/llvm-project/pull/322 in, which should fix the error seen in https://github.com/CHERIoT-Platform/cheri-rust/issues/84, although it won't resolve the underlying issue (i.e. the missing function).

### DIFF
--- a/cheri/tests/format_args.rs
+++ b/cheri/tests/format_args.rs
@@ -1,5 +1,6 @@
 #![no_std]
 
+extern crate alloc;
 extern crate cheriot;
 
 #[no_mangle]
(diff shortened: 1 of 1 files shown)


jacobpake:
> > Should this be squashed or better to keep separate?
> 
> I'd like to keep them separate in this specific case, since they touch two very distant parts of the repository. Does that make sense?

Yes. Do we use "Rebase and merge" button in this case?

PR #123, opened 2026-02-27
Changes required to build for Codasip X730 CHERI CPU

lewis-belsten-revill:
Adds a 64-bit CHERI RISC-V target specifically for Codasip's spark x730 CPU running Musl Linux. Also include necessary changes that fix issues related to CHERI.

For the purpose of this PR we don't update LLVM, since the CHERIoT LLVM changes and Codasip's LLVM changes are not under a unified repository; this PR is just to unify the Rust side before we diverge too much.


compiler/rustc_codegen_llvm/src/debuginfo/metadata.rs
seharris: I'd prefer not to weaken this assertion so much. I'd suggest something like this:
```Rust
// The size and alignment calculations for the vtable assume that a pointer can hold a `usize`, check this is true.
let (usize_size, usize_align) = cx.size_and_align_of(tcx.types.usize);
assert!(pointer_size >= usize_size);
assert!(pointer_align >= usize_align);
```

seharris (commented):
Thanks for the PR, especially the fixes for CHERI stuff! We're definitely interesting in adding more CHERI targets, what do your longer term plans for this look like?
The difference in LLVM fork needed for this target will make CI integration a little tricky, so for the moment we wouldn't have any automated testing. I don't think that's too much of a problem, but it does mean we won't have automatic warning of breakages.
I'm not sure how best to handle the extra dependencies on external repositories, I suspect we'll want to bring those into our existing collection of modified crates at some point to keep things tidy, but I'm not really sure whether that's worth doing right now.

xdoardo:
(Sorry for the troubles that force-pushing to `beta` caused here. It took us a while to figure out CI to correctly push build artefacts, and some commits needed to be squashed.)

lewis-belsten-revill:
> It looks like adding a target feature to `RISCV_FEATURES` (in `compiler/rustc_target/src/target_features.rs`) results in it being passed to LLVM when compiling for any RISC-V target. The comment for the lists of features makes it sound like target features can be used without them being in that list, and also says something about a way to filter out features not supported by older versions of LLVM. From that, it's seems like the solution is to either leave `cap-mode` and `zcheripurecap` out of `RISCV_FEATURES` completely (and just keep the constraints for them), or use `to_llvm_features()` (in `compiler/rustc_codegen_llvm/src/llvm_util.rs`) to make sure they can't get through to LLVM. I've tested that leaving them out fixes building CHERIoT, but I haven't checked whether it will break the new Codasip target.

Sorry it's been a while since I looked at this. I've removed the features from the list but it doesn't build without a fatal error for our target now. Pushed here anyway.

FYI also rebased

seharris:
What kind of error? Does it come from inside the compiler, from LLVM, or from somewhere else?

lewis-belsten-revill:
> What kind of error? Does it come from inside the compiler, from LLVM, or from somewhere else?

Sorry for the delay again, I just re-tried and I have it working without complaint now. I needed to also remove it from the `FeatureConstraints` for the ABI. Will push an update once I've rebased again

lewis-belsten-revill:
Anything left blocking this getting merged?

seharris:
Hopefully that's all of the big things. That leaves the CI failures to clean up, and I want to check with Edoardo how best to handle your forks of `libc` and `cc-rs`. I've had a quick go at fixing the CI stuff, most of the tidy errors are trivial, but there are a couple that look like they'll need more work. In theory I think I should be able to push the trivial fixes to your PR branch, if that's any help?

lewis-belsten-revill:
> Hopefully that's all of the big things. That leaves the CI failures to clean up, and I want to check with Edoardo how best to handle your forks of `libc` and `cc-rs`. I've had a quick go at fixing the CI stuff, most of the tidy errors are trivial, but there are a couple that look like they'll need more work. In theory I think I should be able to push the trivial fixes to your PR branch, if that's any help?

Thanks yes, that'd be helpful. I don't have the CI running locally myself.

seharris:
I'll check my changes and try pushing them once I've sorted out some build problems.

Are the CI results visible to you in this thread? I can't see them when I'm not logged in, but I feel like they should be visible to someone making a pull request. They look like this for me:
screenshot

Either way, the check that's failing is `x.py test tidy` which you should be able to run locally fairly easily. The only problem I've had with it is that it has trouble finding `rustfmt`. My workaround is to build `rustfmt` manually (`x.py build rustfmt`) and add `rustfmt = "./build/host/stage1/bin/rustfmt"` to the `[build]` section of `bootstrap.toml`.

lewis-belsten-revill:
Thanks for the quick fix, I can indeed see the tests, I assume it wants me to add lines in `tests/assembly-llvm/targets/targets-elf.rs` so I can do that. I don't understand what it wants regarding the duplicate deps for `cc` and `libc`?

seharris:
My guess is that something checks for crates being provided by different sources in different parts of the compiler workspace (which `cc` and `libc` are currently). Having different version of the same crate *can* cause problems, I think, so I can see why there might be a check for it. Unfortunately, I'm not really sure what the right solution is, and I'm a bit short on time at the minute so I can't really look into it properly for the moment.

xdoardo:
I just rebased this PR to `beta-new-no-channel`, which will become the new `beta` channel as soon as we can remove the current `beta-old` branch. Just for the record, I think that in the future we will stop with this mechanism of rebasing and branch renaming, so that updates should be a bit smoother.

While rebasing I had to add the definition of the `LlvmAbi`s used originally in the PR, but hopefully changes coming "from me" should stop there.

---
edit: looks like I inadvertently added the (rewinding) changes to the submodules in the previous commit, so I fixed it. Locally the `tidy` tests still fail due to the unknown dependencies on `libc` and `cc`.

Issue #129, opened 2026-03-18
codegen-llvm: failing `instrprof` tests when LLVM has assertions enabled


xdoardo:
```
failures:
    [codegen] tests/codegen-llvm/instrument-coverage/instrument-coverage.rs#all
    [codegen] tests/codegen-llvm/instrument-coverage/instrument-coverage.rs#on
    [codegen] tests/codegen-llvm/instrument-coverage/instrument-coverage.rs#default
    [codegen] tests/codegen-llvm/instrument-coverage/instrument-coverage.rs#y
    [codegen] tests/codegen-llvm/instrument-coverage/instrument-coverage.rs#true_
    [codegen] tests/codegen-llvm/instrument-coverage/instrument-coverage.rs#yes
    [codegen] tests/codegen-llvm/pgo-instrumentation.rs

test result: FAILED. 602 passed; 7 failed; 411 ignored; 0 measured; 0 filtered out; finished in 8.08s
```
 
The first six fail with:
```
Invalid constantexpr cast!: op=50 S = ptr addrspace(200)
Ty = ptr
Assertion failed: (false && "Invalid constantexpr cast!"), function getCast, file Constants.cpp, line 2267.
```
This is due to the `llvm.instrprof.increment` ending up being declared, in the LLVM IR, as
```
declare void @llvm.instrprof.increment(ptr, i64, i32, i32) addrspace(200)
```
while the first argument passed to it when it is called is a `ptr addrspace(200)`. 

The last test fails with 
```
Assertion failed: (V[I]->getType() == ST->getTypeAtIndex(I) && "Initializer for struct element doesn't match!"), function ConstantAggregate, file Constants.cpp, line 1324.
```
which seems to be a similar problem.

As mentioned in the issue title: these all happen only when LLVM assertions are enabled.

PR #130, opened 2026-03-18
Rewrite gen_bootstrap and enable LLVM assertions in CI

xdoardo:
As described in https://github.com/CHERIoT-Platform/cheri-rust/issues/129 enabling the assertions makes some `instrprofile` tests fail, therefore this PR also contains a patch to ignore those tests.


jacobpake (changes requested):
We used to have `profile = "compiler"`, is that something we want to keep as default, or make an option for?

CI tasks are running faster but we might be missing settings like `debug-assertions = true` etc.

It might be that we want to consider some of the `compiler` settings, which are tuned for development rather than CI, e.g. faster rebuilds at the cost of slower initial builds.

(otherwise LGTM and I had fun generating some bootstraps)

jacobpake:
I noticed that the script no longer complains if you have an existing bootstrap.toml and will just overwrite it.

I personally prefer this behaviour, but if we want to avoid this, could we instead add an option (e.g. -f) to allow it?

xdoardo:
[16707a6](https://github.com/CHERIoT-Platform/cheri-rust/pull/130/commits/16707a6d14d59958b6f339ca4471bdb0692cef91) adds the `profile = "compiler"` config back and also adds a `--force` flag that makes the generator overwrite the existing file if there is one.

jacobpake (approved):

cheri/gen_bootstrap.py
```diff
@@ -0,0 +1,237 @@
...
+    ccache: str | bool
+
+    enable_optimize_for_size: bool
+
+    def __repr__(self):
+        return str(self.__dict__)
+
+    def __init__(self, args):
+        self.build_clang = args.build_clang
+        self.use_llvm_config = args.llvm_config_bin is str
```
jacobpake: This is always false (`is` doesn't check type equality, it works for `None` because that is a value/reference)

```suggestion
        self.use_llvm_config = type(args.llvm_config_bin) is str
```

cheri/gen_bootstrap.py
```diff
@@ -0,0 +1,237 @@
...
+            },
+            "rust": {
+                "channel": channel,
+                "std-features": std_features,
+            },
+            "llvm": {"download-ci-llvm": False},
+        }
+
+        if self.use_llvm_config:
+            assert self.llvm_config_bin is str
```
jacobpake:
```suggestion
            assert type(self.llvm_config_bin) is str
```

jacobpake (changes requested):

cheri/gen_bootstrap.py
jacobpake: (alternatively `is not None`)

cheri/gen_bootstrap.py
xdoardo: [018c72d](https://github.com/CHERIoT-Platform/cheri-rust/pull/130/commits/018c72d29dcee80e5e1abc96c2cffe56680083ba) replaced it with `isinstance`, which is closer to what is done in the rest of the script.

cheri/gen_bootstrap.py
xdoardo: [018c72d](https://github.com/CHERIoT-Platform/cheri-rust/pull/130/commits/018c72d29dcee80e5e1abc96c2cffe56680083ba) replaced it with `isinstance`, which is closer to what is done in the rest of the script.

jacobpake (approved):

PR #131, opened 2026-03-18
Run `coretests` in CI

jacobpake:
This PR adds support for running `coretests` in CI.

This includes the following required changes:

- A minimal implementation of `libtest` (alternative is to patch libtest, which does not currently compile).
- Support for `std/alloc` and `std/stdio` in our `facade` target (currently handed off to the RTOS via FFI).
- Splitting the `coretests` library into smaller sets of tests via feature-flags (otherwise we run into size issues).

Future work includes:

- Investigating and documenting failing tests
- Extending the test harness to support more types of test (should_panic, doctests, ...)

Some parts of `test/num` and all of `tests/floats` are not supported. We are missing many libcalls needed here. Better to address that issue first rather than investing/wasting a lot of time to pinpoint and cfg-out each effected test individually.


cheri/ci/library/cheri_test/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        println!("{} ({}:{}) ... ", test.desc.name, test.desc.source_file, test.desc.start_line);
+
+        let Runnable::Test(runnable) = testfn.into_runnable();
+        let result = runnable.run();
+
+        if result.is_err() {
+            // FIXME: handle properly
+            println!("{}", result.unwrap_err());
+            panic!("A test failed!")
```
xdoardo, line 33: Design question: this means that our `libtest` is `--fail-fast` by default. Would it make sense to make it `--no-fail-fast`, so that in CI we can get a complete idea of how many tests are failing following a new change?

cheri/ci/library/cheri_test/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+        } else {
+            pass += 1;
+        }
+    }
+
+    if total != ignored + pass {
+        panic!("Test count mismatch");
+    }
+
+    println!("[OK] total={} ignored={} pass={} ", tests.len(), ignored, pass);
```
xdoardo, line 43: (If we make it `--no-fail-fast` by default, we'd need here a `failed={}` as well)

cheri/ci/library/cheri_test/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    let total = tests.len();
+    let mut ignored = 0;
+    let mut pass = 0;
+
+    println!("running {} tests...", total);
+
+    for test in tests {
+        let TestDescAndFn { desc, testfn } = make_owned_test(test);
+
+        // FIXME: run should_panic tests
```
xdoardo: If I remember correctly we did a short investigation on what it would take to run should-panic tests. I think we mentioned using the SjLj-based mechanism that we use to catch traps which, I imagine, probably doesn't work for `panic!` unless we make it a trap.  Anyways, if I am not misremembering and if you still have some information about it, it could be useful to add the results of that investigation to this comment (or a new issue).

library/std/src/sys/stdio/cheriotrtos.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+pub const STDIN_BUF_SIZE: usize = unsupported_stdio::STDIN_BUF_SIZE;
+
+pub fn is_ebadf(_err: &io::Error) -> bool {
+    true
```
xdoardo, line 65: I see that this is the default from `stdio/unsupported.rs`, but it would be nice to know exactly why `true` and not `false`.

library/coretests/tests/fmt/num.rs
```diff
@@ -266,9 +273,9 @@ fn test_format_int_exp_precision() {
     assert_eq!(format!("{:+10.3e}", 1), "  +1.000e0");
 
     // test precision remains correct when rounding to next power
-    #[cfg(miri)] // can't cover all of `i16` in Miri
+    #[cfg(any(miri, target_abi = "cheriot"))] // can't cover all of `i16` in Miri
```
xdoardo, line 276: Because it's too slow on the simulator, right?

cheri/ci/build/runner.cc
```diff
@@ -0,0 +1,27 @@
...
+  void *ret = heap_allocate(&timeout, MALLOC_CAPABILITY, size);
+  Test::Invariant(CHERI::Capability{ret}.is_valid(),
+                  "Allocation is invalid, got pointer: {} -- {}", ret,
+                  (int)ret);
+  return ret;
+}
+
+extern "C" void cheriot_free(void *ptr) { heap_free(MALLOC_CAPABILITY, ptr); }
+
+extern "C" int __rust_main();
```
xdoardo: I expect it not to be a problem, and I think I've used the same identifier before, but I've been warned enough times to remember that identifiers with two leading underscores in C++ are reserved to the compiler. See, for example, the [C++ 2013 standard](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf#page=444) (17.6.4.3.2 Global names).

cheri/ci/script/coretests.sh
```diff
@@ -0,0 +1,89 @@
...
+ignored_sum=0
+pass_sum=0
+
+for feature in "${features[@]}"; do
+    echo "[RUN] coretests/$feature..."
+    output=$(./cheri/ci/script/coretest.sh $feature 2>&1)
+    results=$(echo "$output" | grep "^\[OK\] ")
+
+    if [[ -z "$results" ]]; then
+        echo "$output"
```
xdoardo: I think that it could be useful to print the full output even if the tests pass, and not only if at least one failed. Sometimes it's handy to be able to make sure that a specific test actually ran at a specific date.

cheri/ci/library/cheri_test/src/types.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+#[derive(Debug)]
+pub struct TestDescAndFn {
+    pub desc: TestDesc,
+    pub testfn: TestFn,
+}
+
+#[inline(never)]
+fn __rust_begin_short_backtrace<T, F: FnOnce() -> T>(f: F) -> T {
```
xdoardo, line 117: Might be worth adding a comment saying that this specific function name is used, on systems that support it, as a "needle" to clean up the backtraces when printing them without `RUST_BACKTRACE=full` (the code implementing this logic is, I think, [here](https://github.com/CHERIoT-Platform/cheri-rust/blob/17d875cb998e242abec27077869d9d2d3e015512/library/std/src/sys/backtrace.rs#L90)). But, of course, we don't support backtraces right now.

cheri/ci/library/cheri_test/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+members = []
```
seharris, line 2: Does this need to be set up as a workspace?

cheri/ci/build/.gitignore
seharris, line 1: I can't see any uses of `compile_flags.txt`, do we still need to ignore that?
Would it be an idea to make the paths here absolute for clarity? (i.e. `/cheri/ci/build/rust_objects.txt`)

library/coretests/tests/net/ip_addr.rs
```diff
@@ -478,6 +479,7 @@ fn ipv4_properties() {
 }
 
 #[test]
+#[cfg(not(target_abi = "cheriot"))] // FIXME: traps, too big?
```
seharris: I think this needs a `FIXME(cheri/triage)` label or to be referenced in an issue otherwise there's a good chance we'll forget about this test.

cheri/ci/build/xmake.lua
seharris, line 1: I'd suggest adding comments to the top of this file and the shell scripts to explain what all these scripts are for, because there are quite a few moving parts and it's not super easy to figure out what's going on, and which files are related. Something like `// Builds firmware for a module of the coretests suite, see cheri/ci/script/coretests.sh for more details` and then maybe put a more detailed explanation in `coretests.sh` or something.

library/coretests/tests/lib.rs
```diff
@@ -163,56 +169,122 @@ macro_rules! test_runtime_and_compiletime {
...
 mod const_ptr;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_convert"))]
 mod convert;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_ffi"))]
 mod ffi;
+#[cfg(not(target_abi = "cheriot"))] // FIXME(cheri/triage): needs work, libcalls + ???
 mod floats;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_fmt"))]
 mod fmt;
+#[cfg(all(target_abi = "cheriot", feature = "test_fmt_builders"))]
```
seharris, line 210: This is a nitpick, but why are these here instead of `library/coretests/tests/fmt/mod.rs` where they're referenced on other targets?

library/coretests/tests/lib.rs
```diff
@@ -163,56 +169,122 @@ macro_rules! test_runtime_and_compiletime {
...
 mod hint;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_index"))]
 mod index;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_intrinsics"))]
 mod intrinsics;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_io"))]
 mod io;
+#[cfg(any(not(target_abi = "cheriot"), feature = "test_iter"))]
 mod iter;
+#[cfg(all(target_abi = "cheriot", feature = "test_iter_adapters"))]
```
seharris, line 233: Same thing as my comment about the `fmt/*` declarations earlier in this file (i.e. why not put this in `library/coretests/tests/iter/mod.rs`?)

cheri/ci/script/coretest.sh
seharris, line 1: Might be an idea to add `|| exit 1` to handle errors in the git and xmake commands, or `set -e` if that won't interfere with how you're using Cargo (and I can't remember if it breaks `[[ -d ]]`).

seharris (commented):
Edoardo beat me to it, but I may as well add my comments now I've typed them. They're just minor suggestions.

cheri/ci/library/cheri_test/src/lib.rs
jacobpake: I agree that should be the default approach, if it causes any issues we can switch back.

This will need panic handling (future work, #144).

jacobpake: This particular snippet handles non-panic failures (tests may return with `Result`, we only have one such failure currently).

cheri/ci/library/cheri_test/src/lib.rs
jacobpake: Yes, I think it merits a separate discussion, so will not implement here, logged in #144

cheri/ci/library/cheri_test/Cargo.toml
jacobpake: Yes, otherwise:

> error: current package believes it's in a workspace when it's not:
> current:   /Users/jacob/cheriot/rust/cheri/ci/library/cheri_test/Cargo.toml
> workspace: /Users/jacob/cheriot/rust/Cargo.toml
> 
> this may be fixable by adding `cheri/ci/library/cheri_test` to the `workspace.members` array of the manifest located at: /Users/jacob/cheriot/rust/Cargo.toml
> Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.

cheri/ci/build/.gitignore
jacobpake: > I can't see any uses of compile_flags.txt

I use one here to tell my ide/lsp about cheri.

> Would it be an idea to make the paths here absolute for clarity? (i.e. /cheri/ci/build/rust_objects.txt)

I think it needs to be relative to the .gitignore, unless you mean to add to the top-level .gitignore?

seharris: Oh duh, I hadn't noticed that wasn't the top level one! Oops.

library/coretests/tests/lib.rs
jacobpake: The problem is that these modules need to be further split up, but the entrypoint also contains tests as well as the import statements. This meant in order to avoid running duplicate tests, we would need to add #[cfg]'s over all the other tests/helpers/macros (simultaneously battling with unused warnings in some cases).

This was especially apparent in `num`, where the entry point contains ~700 lines of tests.

I decided that doing it this way would reduce the surface area of the change and be easier to maintain. If we prefer an alternative let me know!!

cheri/ci/script/coretest.sh
jacobpake: I usually prefer to use `set -e` by default, I'm not sure why I omitted it here, I'll check! I don't think it should intefere with anything...

library/std/src/sys/stdio/cheriotrtos.rs
jacobpake: Good question. As far as I can tell it is only used by [handle_ebadf](https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/std/src/io/stdio.rs#L203), which is only used in the `StdoutRaw`/`StderrRaw` impls above.

There is some related discussion here: https://github.com/rust-lang/rust/issues/47271.

jacobpake: See also: https://github.com/rust-lang/rust/issues/85121

jacobpake: And: https://github.com/retep998/rfcs/blob/println/0000-stdout-existential-crisis.md

cheri/ci/build/runner.cc
jacobpake: Changed to `rust_main`

library/coretests/tests/net/ip_addr.rs
jacobpake: nice spot, fixed in 150248e98ab97ee6231c4c08b99c8244d2a23b02

cheri/ci/build/xmake.lua
jacobpake: I've moved everything referencing the build directory into the linker script, removed the mostly redundant runner script, and added some documentation at the the top of each remaining script.

If you think it needs more clarification, I could change the structure/naming so the hierarchy is more apparent:

```
cheri/ci
└── coretests
    ├── build
    │   ├── runner.cc
    │   └── xmake.lua
    ├── linker.sh
    ├── run-all.sh
    └── run.sh
```

library/coretests/tests/fmt/num.rs
jacobpake: yes, good spot, labelled in 4311190ef674bde5f162bf886e7a562d36589987

cheri/ci/script/coretests.sh
jacobpake: We print the full output in 4311190ef674bde5f162bf886e7a562d36589987

cheri/ci/script/coretest.sh
jacobpake: These commands are moved to `linker.sh` with `set -e` in 4311190ef674bde5f162bf886e7a562d36589987

cheri/ci/library/cheri_test/src/types.rs
jacobpake: Added your comment near verbatim in [f310392](https://github.com/CHERIoT-Platform/cheri-rust/pull/131/commits/f310392361e81bd601fa5b970edc104ec5645912)

library/coretests/tests/fmt/num.rs
```diff
@@ -237,6 +243,7 @@ fn test_format_int_exp_limits() {
 }
 
 #[test]
+// #[cfg(not(target_abi = "cheriot"))] // hanging
```
seharris: Did you mean to commit this commented conditional?

cheri/ci/build/xmake.lua
seharris: I think the changes you've already made work ok as they are.

library/coretests/tests/lib.rs
seharris: Ok, yeah I see. I think you've chosen the right trade off then.

xdoardo (approved):

library/coretests/tests/fmt/num.rs
jacobpake: fixed!

seharris (approved):

PR #132, opened 2026-03-20
wip: ci caching experiment

jacobpake:
This PR:

- Caches LLVM and Sail sim builds
- Separates `x check` and `x test` into distinct tasks
- Runs all other tasks in parallel after `x check` has passed
- Some other minor changes

**Caching builds**

Currently, the LLVM cache key is a hash of:

- host target triple
- bootstrap.toml contents
- LLVM commit SHA

We have discussed how to further scope build caches. IMO there are three options:

1. Shared between all PRs
2. Shared between all runs in a single PR
3. Shared between all tasks in a single run

Ordered by safety (ascending) and effectiveness (descending).

Normally, bootstrap will also do some fingerprinting, but I am currently bypassing this as it wasn't playing well with (shallow clone + CI + submodules). In order to consider option 1 we should have a solution to this.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,88 +9,151 @@ gce_instance:
   disk: 60
   spot: true
 
-x_test_task:
-  name: Run ./x test
-  alias: x_test
-  timeout_in: 240m
-  dependencies_script:
-    - set -eo pipefail
+env:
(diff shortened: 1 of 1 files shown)


jacobpake:
No longer relevant as Cirrus is shutting down.

PR #137, opened 2026-03-31
Bump LLVM to f31a26ad9785a and add ptrtoint test (and test CI)

xdoardo:
This LLVM commit fixes two bugs that arose when compiling Rust code:
1. Lowering of ptrtoint addrspace(200), i64
2. Lowering of frem double

---

Piggybacking on this PR a patch to the runner to explicitly fail if a test fails.

We used this PR to test how our CI setup behaves when the incoming branch is behind the base branch, specifically in the .cirrus.yml setup. TLDR: pay attention when GitHub tells you that the branch is outdated. 

Original PR: https://github.com/CHERIoT-Platform/cheri-rust/pull/134.

### DIFF
--- a/cheri/tests/ptrtoint.rs
+++ b/cheri/tests/ptrtoint.rs
@@ -1,0 +1,14 @@
+#![no_std]
+
+extern crate cheriot;
+
+#[no_mangle]
+extern "C" fn test_ptrtoint() -> i32 {
+    if core::hint::black_box(do_it()) as i32 == 0 { 1 } else { 0 }
+}
+
+#[inline(never)]
+fn do_it() -> i64 {
+    let x = 0;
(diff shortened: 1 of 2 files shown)


cheri/tests/ptrtoint.rs
```diff
@@ -0,0 +1,14 @@
+#![no_std]
+
+extern crate cheriot;
+
+#[no_mangle]
+extern "C" fn test_ptrtoint() -> i32 {
+    if core::hint::black_box(do_it()) as i32 == 0 { 1 } else { 0 }
```
jacobpake, line 7: this isn't the main failure case but to ensure the compiler doesn't optimise away (note to self)

jacobpake (approved):

PR #138, opened 2026-04-01
Add `target_address_width` attribute and `address_size` field within target specification

seharris:
This PR adds a new attribute ~~`target_usize_width`~~ `target_address_width` and replaces cases in `core`, `std`, and parts of the test suite relevant to us where `target_pointer_width` is used to get the size of `usize`.

~~Unfortunately, I couldn't find an easy way to avoid needing to add `pointer_offset` to the target specification, which means updating every target and making the diff for the PR quite large. It's still possible that there is a way to avoid this, but it will take more time to investigate and I wanted to get this work to state where I could get opinions on it faster. I'm making this a draft until we've decided what to do about this.~~

Update: while these changes require adding `address_size` (used to be `pointer_offset`) to the target specification, we can make it default to being the same as the pointer size, such that it only needs to be set on CHERI targets. This keeps the change quite a lot smaller.

Adding ~~`pointer_offset`~~ `address_size` is, however, potentially useful because it's most of the work needed to make compile test correctly (I think?) classify CHERIoT as a 32 bit target, if we want that. (The rest of the changes needed are outside the scope of this PR) I've already tested implementing this, and it works but didn't yield the extra test coverage from the `codegen-llvm` suite that I was hoping for. If this is true of the other test suites then it may not be worth the time.

### DIFF
--- a/compiler/rustc_target/src/spec/mod.rs
+++ b/compiler/rustc_target/src/spec/mod.rs
@@ -2194,6 +2194,16 @@ impl Target {
             });
         }
 
+        if let Some(target_address_width) = self.options.address_width {
+            let dl_address_size: u64 = dl.address_size().bits();
+            if dl_address_size != Into::<u64>::into(target_address_width) {
+                return Err(TargetDataLayoutError::InconsistentTargetAddressWidth {
+                    address_size: dl_address_size,
+                    target: target_address_width,
+                });
+            }
+        }
+
         dl.c_enum_min_size = Integer::from_size(Size::from_bits(
             self.c_enum_min_bits.unwrap_or(self.c_int_width as _),
         ))
@@ -2270,6 +2280,10 @@ type StaticCow<T> = Cow<'static, T>;
 pub struct TargetOptions {
     /// Used as the `target_endian` `cfg` variable. Defaults to little endian.
     pub endian: Endian,
+    /// Number of bits in the address part of a pointer, a `usize`, and the machine word size.
+    /// This decides the value of the `target_address_width` `cfg` variable.
+    /// Defaults to `pointer_width`.
+    pub address_width: Option<u16>,
     /// Width of c_int type. Defaults to "32".
     pub c_int_width: u16,
     /// OS name to use for conditional compilation (`target_os`). Defaults to [`Os::None`].
@@ -2800,6 +2814,7 @@ impl Default for TargetOptions {
     fn default() -> TargetOptions {
         TargetOptions {
             endian: Endian::Little,
+            address_width: None,
             c_int_width: 32,
             os: Os::None,
             env: Env::Unspecified,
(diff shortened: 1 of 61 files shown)


jacobpake:
> Unfortunately, I couldn't find an easy way to avoid needing to add pointer_offset to the target specification, which means updating every target and making the diff for the PR quite large. It's still possible that there is a way to avoid this, but it will take more time to investigate and I wanted to get this work to state where I could get opinions on it faster. I'm making this a draft until we've decided what to do about this.

The only way I can think of is to move `pointer_offset` under `TargetOptions` (which apparently exists to support defaults), and handle as `max_atomic_width` is handled:

```rs
impl Target {
    pub fn max_atomic_width(&self) -> u64 {
        self.max_atomic_width.unwrap_or_else(|| self.pointer_width.into())
    }
}

pub struct TargetOptions {
    /// Don't use this field; instead use the `.max_atomic_width()` method.
    pub max_atomic_width: Option<u64>,
}

impl Default for TargetOptions {
    fn default() -> TargetOptions {
        TargetOptions {
            max_atomic_width: None,
        }
    }
}
```

seharris:
Oh yeah, that's a good idea. I'll have a look at that when I get a moment and try and figure out if that's the way to do this.

seharris:
(Rebase on to beta in progress)

compiler/rustc_target/src/spec/mod.rs
```diff
@@ -2203,6 +2213,10 @@ type StaticCow<T> = Cow<'static, T>;
 pub struct TargetOptions {
     /// Used as the `target_endian` `cfg` variable. Defaults to little endian.
     pub endian: Endian,
+    /// Number of bits in the address part of a pointer, a `usize`, and the machine word size.
+    /// This decides the value of the `target_usize_width` `cfg` variable, and whether compiletest thinks the target is 32 or 64 bit.
```
lygstate: target_address_width

lygstate:
The title may also need to be target_address_width

compiler/rustc_target/src/spec/mod.rs
xdoardo: I sort of agree with the suggestion, but it would be nice to know why you think the current name should be changed.

lygstate: The author already changed, just not everywhere.

xdoardo: Ah, I missed that, sorry.

lygstate: > I sort of agree with the suggestion, but it would be nice to know why you think the current name should be changed.

I think `target_address_width` better, because we also have `isize`, `target_usize_width` is too specific

xdoardo: I agree. It's also more conformant with the nomenclature we use in other places (see, for example, #153). I was asking because since I hadn't reviewed this PR yet I didn't know the name was changed already, and having more information about the suggestion would've been useful. Again, I agree and thank you for the input.

seharris: Good catch, thanks!

seharris:
This is now split into two commits to make things easier to follow, and excludes a change to `std` that wouldn't be tested. There are still a number of changes to `std`, in one case because I'm really sure it's correct (`library/std/src/io/cursor/tests.rs`), and another because it's required to build the facade target (`library/std/src/io/error.rs`). There are also still a few changes to other libraries and tests like `portable-simd`, `alloctests`, and `compiler-builtins` because I think we're either using them, testing them, or will use and/or test them. Let me know if you disagree with my call on any of these, I might be wrong!

xdoardo:
These changes also make all tests in coretests/slice pass that we currently ignore due to tag violations pass. Nice!

Edit: also the test in `coretests/convert.rs`.

xdoardo (approved):

PR #143, opened 2026-04-15
Delete upstream funding, issue template, and workflow metadata to avoid confusion

seharris:
This change removes the donation box on the Github landing page, ~~upstream CI configuration we aren't using,~~ and templates for issues and pull requests which I don't think have much relevance to how we work. I'm hoping this will reduce the chances of confusing anyone who wants to submit bug reports, and make sure we can't accidentally use the wrong CI options.


jacobpake:
I agree with the principle of this PR but if it requires any changes to bootstrap/internal testing then removing those files will probably not be a net positive.

(Surprised again by the pervasive test coverage here!!)

seharris:
Wow, yup, really wasn't expecting our CI to notice this change. I'll see if there's a way to do it without messing with testing, possibly just leaving the CI files as they were and removing everything else is the right way to do this.

jacobpake (commented):
I'm happy with this unless @xdoardo has a problem.

xdoardo (approved):

xdoardo:
Hopefully it's not too much of a pain to rebase this on top of `beta`.

PR #146, opened 2026-04-22
cheri/runner: correctly handle tests in subdirectories

xdoardo:
Because `ancestors()` returns entire subpaths, not just the components of the path. 

[Permalink to rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=49e560f9f39805d7ec9df077fe8b2013).


jacobpake (approved):
works in my testing and the code reads better

PR #148, opened 2026-04-24
Changes to allow CHERI capability registers in inline asm

lewis-belsten-revill:


seharris:
Thanks for the pull request!

Do you have any tests for this that could be adapted into something we can add to the test suite? It'd be nice to have a smoke test or something for this feature.

Have you checked what happens if someone tries to use capability registers on a non-CHERI target? I think ideally we want to either make sure they're only recognised on CHERI targets, or that the compiler produces a sensible error when they're used on non-CHERI targets.

lewis-belsten-revill:
Good point, should be simple to add some ability to disable these registers on non-CHERI. I will also add tests

xdoardo:
Rebased on top of `beta-new-no-channel` and fixed the `tidy` errors. 

--- 
edit: But not the errors coming from `rustc_codegen_gcc`!

PR #147, opened 2026-04-22
Bump LLVM to 94de516, fix LSR-induced bugs, add `+relax` feature

xdoardo:
This PR should fix #139, #140 and #141. 

As mentioned in those issues, those errors were caused by LSR-induced bugs, and [this PR to LLVM ](https://github.com/CHERIoT-Platform/llvm-project/pull/365) fixes the LSR pass so that it doesn't generate faulty code anymore. Updating to this new LLVM version introduced new regressions, which are now mostly fixed (with other patches to LLVM).

This PR also adds the `+relax` feature to our target, which was part of the mandatory configuration, which we never enabled.

----

Requires https://github.com/CHERIoT-Platform/cheri-rust/pull/150 to be merged first to fix the `xmake` rules generated by the test runner.

PR #151, opened 2026-05-06
rustc_const_eval: Fix computation of `TypeId`

xdoardo:
The CTFE machinery was using `pointer_size()` where it should have been `pointer_offset()`. This PR removes the "ignore" annotations on tests that were failing due to this error.

----

This PR depends on https://github.com/CHERIoT-Platform/cheri-rust/pull/147.

### DIFF
--- a/compiler/rustc_const_eval/src/const_eval/type_info.rs
+++ b/compiler/rustc_const_eval/src/const_eval/type_info.rs
@@ -77,7 +77,7 @@ impl<'tcx> InterpCx<'tcx, CompileTimeMachine<'tcx>> {
         // Fill all fields of the `TypeInfo` struct.
         for (idx, field) in ty_struct.fields.iter_enumerated() {
             let field_dest = self.project_field(dest, idx)?;
-            let ptr_bit_width = || self.tcx.data_layout.pointer_size().bits();
+            let address_bit_width = || self.tcx.data_layout.pointer_offset().bits();
             match field.name {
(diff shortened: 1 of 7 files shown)


compiler/rustc_const_eval/src/const_eval/type_info.rs
seharris, line 1: I'd suggest renaming `ptr_bit_width` to `address_width` or something, because this is a bit confusing currently.

library/coretests/tests/mem.rs
seharris, line 1: `#[cfg()]` works on single lines too, so the assertions here don't need to be in bracketed blocks if you think that would be tidier. You can just do
```Rust
#[cfg(target_family = "cheriot")]
assert_eq!(size_of::<*const usize>(), 8);
```

xdoardo: [97ff5f9](https://github.com/CHERIoT-Platform/cheri-rust/pull/151/commits/97ff5f996e9b1042a2334e53ab4b14fd7ba711d8)

compiler/rustc_const_eval/src/const_eval/type_info.rs
xdoardo: [6833862](https://github.com/CHERIoT-Platform/cheri-rust/pull/151/commits/6833862a881f35ce55da551db6236b2d1f816599)

seharris (approved):

PR #154, opened 2026-05-12
cheri/runner: handle bare filename in lca guard

### LINKED ISSUE #37: The runner does not work with single test files
As per title: 
```
$ runner/target/release/cheriot-runner smoke.rs --rustc-toolchain=cheri

thread 'main' (6761526) panicked at src/runner/build.rs:62:22:
test file parent should have a valid file name
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

christiankissig:
- Fixes #37. Passing a bare filename like `smoke.rs` to `cheriot-runner` no longer panics.
- The parent of `smoke.rs` is empty: `""`. `""` has not been handled, so that bare filenames with empty parent fell through and caused a panice in `file_name().expect(...)`. This change is adding `""` to the guard.
- The bare filename `smoke.rs` is handled like `./smoke.rs`.
- Change has been tested in containerised environement.

### DIFF
--- a/cheri/tests/runner/src/runner/build.rs
+++ b/cheri/tests/runner/src/runner/build.rs
@@ -38,7 +38,10 @@ impl App {
             // This is definitely not the right way to compute the LCA of test_path w.r.t. the cwd,
             // but we'll work with it for now.
 
-            if test_parent_path.as_os_str() == "." || test_parent_path.as_os_str() == ".." {
+            if test_parent_path.as_os_str().is_empty()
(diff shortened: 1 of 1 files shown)


xdoardo:
Rebased on `beta` and changed the target branch to that, so that we can directly merge there without cherry-picking afterwards.

xdoardo (approved):

xdoardo:
The error in CI is unrelated to this PR. Merging.

PR #155, opened 2026-05-12
cheri/libtest: exception handling

jacobpake:
This PR adds exception handling to our minimal libtest. This allows us to run `should_panic` tests (previously ignored) and to have `no-fail-fast` behaviour i.e. a test failure does not cause the suite to exit.

Panics from Rust, CHERI exceptions, and other errors during test execution should be caught and unwound via the Setjmp/Longjmp implementation [in the RTOS](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/include/unwind.h). The implementation on the Rust side is mostly copied from [this demo](https://github.com/SCISemi/devex-demos/blob/main/rust/ffi/ffi/src/cheriot.rs).

When compiling with the standard library, we cannot define a `panic_handler` as we do in `no_std`, so we have to set a panic hook to capture panic information. The implementation is not ideal, and in the future, especially if we want to support the std library in any form, we will need to either define a new `panic_runtime` or adapt `panic_abort`/`panic_unwind` for the RTOS.

The upstream `libtest` uses threads if the panic runtime is panic_abort. This might be something that we can support via the RTOS as an alternative to the approach here.

### DIFF
--- a/cheri/ci/build/runner.cc
+++ b/cheri/ci/build/runner.cc
@@ -1,27 +1,75 @@
 #define MALLOC_QUOTA 0x100000
 
-#include "fail-simulator-on-error.h"
 #include <allocator.h>
+#include <compartment.h>
 #include <debug.hh>
+#include <priv/riscv.h>
+#include <simulator.h>
+#include <unwind.h>
 
-using Test = ConditionalDebug<true, "Test Runner">;
+using Debug = ConditionalDebug<true, "Test Runner">;
+
+// Display information about a CHERI fault, then either unwind or exit.
+// Aborts from Rust should trigger an invalid instruction which is caught here.
+extern "C" ErrorRecoveryBehaviour
+compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval) {
+  if (mcause == priv::MCAUSE_CHERI) {
+    // Note: handle CZR differently as `get_register_value` will return a
+    // nullptr which we cannot dereference.
+
+    auto [exceptionCode, registerNumber] = CHERI::extract_cheri_mtval(mtval);
+
+    Debug::log("{} error at {} (return address: {}), with capability register "
+               "{}: {}",
+               exceptionCode, frame->pcc,
+               frame->get_register_value<CHERI::RegisterNumber::CRA>(),
+               registerNumber,
+               registerNumber == CHERI::RegisterNumber::CZR
+                   ? nullptr
+                   : *frame->get_register_value(registerNumber));
+  }
+
+  // Calling cleanup_unwind without a registered handler is dangerous,
+  // and if we don't have a handler then we want to ensure we exit with
+  // a failure status (returning ForceUnwind will exit with success).
+  struct CleanupList **head = cleanup_list_head();
+  bool has_cleanup_handler = *head != nullptr;
+
+  if (has_cleanup_handler) {
+    Debug::log("Error, unwinding...");
+    cleanup_unwind();
+  } else {
+    Debug::log("Error, exiting...");
+    simulation_exit(1);
+  }
+
+  // if either call above fails (e.g. because we are not in a simulator or
+  // there is no stack space available) fallback to forcibly unwinding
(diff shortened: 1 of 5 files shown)


cheri/ci/build/runner.cc
```diff
@@ -1,27 +1,74 @@
...
+
+  if (has_cleanup_handler) {
+    Debug::log("Error, unwinding...");
+    cleanup_unwind();
+  } else {
+    Debug::log("Error, exiting...");
+    simulation_exit(1);
+  }
+
+  // Unreachable (?)
```
seharris: Would it be sensible to add some sort of assertion or exit here so that it will be visible if our assumption that this is unreachable is violated?

cheri/ci/script/coretests.sh
```diff
@@ -81,26 +82,32 @@ for feature in "${features[@]}"; do
 
     if [[ $status -ne 0 ]]; then
         echo "Simulator exited with failure"
-        exit 1
     fi
 
-    # e.g. [OK] total=6 ignored=0 pass=6
-    results=$(echo "$output" | grep "^\[OK\] ")
+    # e.g. [OK] total=6 ignored=0 pass=6 fail=0
+    results=$(echo "$output" | grep "^\[OK\|FAIL\] ")
```
christiankissig: GNU grep binds \[ tighter than \|, so that the pattern is equivalent to (^\[OK)  \|  (FAIL\]) . I think you want ^\[OK\]\|^\[FAIL\] or ^\[(OK|FAIL)\] here?

cheri/ci/library/cheri_test/src/panic.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+thread_local! {
+    static PANIC_MESSAGE: RefCell<String> = RefCell::new(String::new());
+    static PANIC_EXPECT: Cell<bool> = const { Cell::new(false) };
+}
+
+pub(crate) fn set_panic_hook() {
+    let prev = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |info| {
```
christiankissig, line 49: The std::panic hook is process global https://doc.rust-lang.org/std/panic/fn.set_hook.html , but PANIC_EXPECT is thread local. Probably ok as long as we don't expect the test to spawn a child thread. Might be worth adding a comment though.

christiankissig (changes requested):

cheri/ci/script/coretests.sh
jacobpake: Fixed with https://github.com/CHERIoT-Platform/cheri-rust/pull/155/changes/34625fb54acd0a6103169841a8ec16b364386b2f

cheri/ci/library/cheri_test/src/panic.rs
jacobpake: We don't have support for threads at the moment, if/when we do we would likely adapt the test runner to use a thread per test which would ideally make this panic handling redundant.

Regardless, I think it makes sense for this to be thread local (if PANIC_EXPECT is not set, the hook calls the default hook).

cheri/ci/build/runner.cc
jacobpake: This was a note/question for myself I should have cleaned up. It is technically reachable (e.g. if we are not in a simulator) but in that circumstance something has gone very wrong already and returning `ForceUnwind` is probably sensible.

I have changed the comment in https://github.com/CHERIoT-Platform/cheri-rust/pull/155/commits/b911f3d1eb54f37190fd9bf348a7f759af279fb7

cheri/ci/library/cheri_test/src/panic.rs
christiankissig: I suppose the global hook would be the first place to look if the thread local flag was not respected, as explicitly made thread local. A comment would be preferable, but i can live with this.

christiankissig (approved):
LGTM

jacobpake:
rebased on beta ahead of merge

xdoardo (approved):

PR #158, opened 2026-05-15
coretests: add missing float tests

jacobpake:
Follows https://github.com/CHERIoT-Platform/cheri-rust/pull/231.

~~This PR uses a fork of the RTOS containing the freebsd/msun implementations for a number of missing float libcalls.~~

This PR adds some stubs for missing libcalls that turn the linktime errors into runtime errors.

This allows us to run the previously absent coretest "submodules":

- `num/floats`
- `num/float_ieee754_flt2dec_dec2flt`

This PR reorganises how `num` is divided in a way that allows the various submodules to share dependencies. This resulted in additional `cfg` attributes, but this should be easier to maintain going forwards.

We also include the (relatively new) `num/carryless_mul` submodule under the `test_num` feature.


cheri/ci/build/runner.cc
```diff
@@ -48,7 +48,12 @@ compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval) {
   return ErrorRecoveryBehaviour::ForceUnwind;
 }
 
-extern "C" void cheriot_print(char *s) { printf("%s", s); }
+extern "C" void cheriot_print(char *str) {
+    for (; *str; ++str) {
+        char c = *str;
+        MMIO_CAPABILITY(Uart, uart)->blocking_write(c);
+	}
+}
```
xdoardo, line 56: Cool, makes sense. I wonder what error brought you to this change :p

jacobpake: Ha, the official reason is to reduce dependencies (printf is provided by library/stdio), but I did do this when I was grasping at straws trying to ensure there were no hidden float libcalls.

jacobpake: (this is what printf is doing in the end, we don't need any of the formatting stuff)

cheri/ci/build/xmake.lua
```diff
@@ -5,8 +5,14 @@ set_toolchains("cheriot-clang")
...
     set_default("sail")
 
+library("stubs")
+	add_cxflags("-include " .. "stubs.h")
+    add_files("stubs.cc")
+
 compartment("test_runner")
-    add_deps("atomic","debug","freestanding","softfloat","stdio")
+    add_deps("atomic", "debug", "freestanding", "softfloat")
+    add_deps("softfloat32pow", "softfloat64pow")
```
xdoardo, line 14: In case you wanted to include all `softfloat` libs, there is a `softfloatall` dependency that adds `...pow` and `...complex` libraries as well.

xdoardo (approved):

cheri/ci/build/xmake.lua
jacobpake: Yes - this was deliberate! We don't need `complex` and in a couple of modules `softfloatall` causes space issues

PR #159, opened 2026-05-15
runner: print captured stdout when end marker is missing

### LINKED ISSUE #38: The runner does not report the message of panics
Take, for example, the following test: 
```
#![no_std]

extern crate alloc;
extern crate cheriot;

#[no_mangle]
extern "C" fn test_panic() -> i32 {
    assert!(false, "hello!!!");
    0
}
```
when executing it the runner will correctly pick up the failure, but won't show the full message: 
```
[1/4] test panic ... fail
--- begin stdout from test runner --
Board file saved as  build/cheriot/cheriot/release/test.board.json
Remapping priority of thread 1 from 1 to 0
@rust-test-runner-sync-start

--- end stdout from test runner --
--- begin stderr from test runner --
tohost located at 0x80014fc8

--- end stderr from test runner --
running `xmake`
running cd "/var/folders/yk/spflkwdd62xd8zhy_nx9_6nr0000gn/T/cheri-runner/objs/unsafecell" && "xmake" "run"
running `xmake run`...ok
```

Instead running the firmware directly (without using the runner as intermediary) will print: 
```
$ xmake run
Board file saved as  build/cheriot/cheriot/release/test.board.json
Remapping priority of thread 1 from 1 to 0
Running file test.
ELF Entry @ 0x80000000
tohost located at 0x80014fc8
@rust-test-runner-sync-start
panicked at ./panic.rs:8:5:
hello!!!FAILURE: 1
error: execv(cheriot_sim test) failed(1)
warning: config.buildir() has been deprecated, please use config.builddir()
warning: add -v for getting more warnings ..
```

christiankissig:
* Fixes #38. Prints `stdout` also when end marker is not found.
* A test that panics inside the firmware never returns from `run_tests`, so the C++ shim's end marker is never printed. This change captures `stdout` also in case, the end marker is not found. `stdout` is then captured unabridged. The start marker remains required.
* Printing still terminates, as the child process has terminated and the buffer is fully materialised at this point.

### DIFF
--- a/cheri/tests/runner/src/runner/mod.rs
+++ b/cheri/tests/runner/src/runner/mod.rs
@@ -174,16 +174,16 @@ impl App {
                     let mut stdout = stdout.split_off(prefix + start_needle.len());
                     if let Some(suffix) = stdout.rfind(good_end_needle) {
                         _ = stdout.split_off(suffix);
+                    }
 
-                        if stdout.is_empty() {
-                            infoln!(self, "");
-                        } else {
-                            println!("-- begin stdout from test {name}");
-                            print!("{stdout}");
-                            println!(
-                                "\n-- end stdout from test {name} (run with more verbosity to see the entire output)"
-                            );
-                        }
+                    if stdout.is_empty() {
+                        infoln!(self, "");
+                    } else {
+                        println!("-- begin stdout from test {name}");
+                        print!("{stdout}");
+                        println!(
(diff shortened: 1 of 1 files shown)


xdoardo (approved):

xdoardo:
Note: currently our CI set up doesn't like when we switch target branches, and for some reason the job isn't picked up when the target branch changes.

PR #160, opened 2026-05-18
ci: migrate CI workflow to GitHub Actions

jacobpake:
Cirrus is discontinuing their service. This PR migrates our CI workflow to the GitHub Actions workflow syntax. We use Cirun as the runner provider, which is configured to use our GCP for compute. Cirun is not integral to this setup and can be switched out for a different runner provider.

Notable differences in behaviour:

- We no longer build the distribution artifacts here. These were only used by the devcontainer, and in the future the devcontainer will build and install Rust directly instead.
- The three other jobs are merged into a single job. This reduces our exposure to Cirun's state management (in testing, there were instances of jobs not being picked up). As a side effect, the overall runtime should be reduced as we are no longer repeating work.
- Removed ccache, a future PR will reinstate this.
- Move `x test tidy` ahead of `x check`
- Adds and makes use of two bootstrap config files, which do not use the `compiler` profile (I think it is an inappropriate profile to use in CI), and are more straightforward to modify whilst debugging/experimenting vs modifying the gen_bootstrap program.

Any further desired changes to behaviour (e.g. optimisations, QoL improvements) to be addressed after this is merged.

TODO (after merge):
- [x] Raise a PR removing our Cirrus config.

### DIFF
--- a/.cirun.yml
+++ b/.cirun.yml
@@ -1,0 +1,12 @@
+runners:
+  - name: cirun-gcp-aarch64-runner
+    cloud: gcp
+    instance_type: n4a-standard-32
+    machine_image: ubuntu-minimal-2404-noble-arm64-v20260507
+    region: us-east1-b
+    preemptible: true
+    extra_config:
+      baseDisk:
+        diskSizeGb: 120
+    labels:
+      - cirun-gcp-aarch64
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -1,0 +1,116 @@
+name: CI (CHERIoT)
+
+on:
+  push:
+    branches:
+      - "beta"
+  pull_request:
+    branches:
+      - "**"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: Run tests
+    runs-on: cirun-gcp-aarch64--${{ github.run_id }}
+    timeout-minutes: 240
+    env:
+      CC: "clang"
+      CXX: "clang++"
+    steps:
+      - name: Install dependencies
+        run: |
+          set -eo pipefail
+          sudo apt update
+          sudo apt install -y pkg-config clang ninja-build lld cmake perl opam z3 libgmp-dev libssl-dev git
+
+        # after apt install git
+      - uses: actions/checkout@v5
+        name: Checkout
+        with:
+          fetch-depth: 2
+          # TODO: src/ci/scripts/checkout-submodules.sh should be faster
+          submodules: recursive
+
+      - name: Configure bootstrap
+        run: cp cheri/bootstrap/ci.toml bootstrap.toml
+
+      - name: Build LLVM
+        run: ./x build llvm --ci=false
+
+      - name: Build rustfmt
+        run: ./x build rustfmt --ci=false
+
+      - name: Run `x test tidy` (host)
+        run: ./x test tidy --ci=false
+
+      - name: Run `x check` (host)
+        run: ./x check --ci=false
+
+      - name: Run `x test` (host)
+        run: ./x test --skip src/tools/linkchecker --skip tests/rustdoc-ui --skip tests/run-make --skip tidy --ci=false
+
+      - name: Build rustc, core (CHERIoT)
+        run: ./x build compiler std --target=riscv32cheriot-unknown-cheriotrtos --ci=false
+
+      - name: Run `x test tests/codegen-llvm` (CHERIoT)
+        run: ./x test tests/codegen-llvm --target=riscv32cheriot-unknown-cheriotrtos --skip src/tools/linkchecker --ci=false
+
+      - name: Install xmake
+        run: |
+          curl -fsSL https://xmake.io/shget.text | bash
+          echo "PATH=$HOME/.local/bin:$PATH" >> $GITHUB_ENV
+
+      - name: Check xmake
+        run: xmake --version
+
+      - name: Install Rust (upstream stable)
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
+          . "$HOME/.cargo/env" && echo "PATH=$PATH" >> $GITHUB_ENV
+
+      - name: Check rust
+        run: cargo --version
+
+      - name: Build Sail simulator
+        run: |
+          git clone --recurse-submodules --depth=1 https://github.com/CHERIoT-Platform/cheriot-sail
+          opam init --reinit --bypass-check --disable-sandboxing --disable-shell-hook
+          eval $(opam env)
+          opam install sail --confirm-level=unsafe-yes
+          cd cheriot-sail && make csim && sudo cp c_emulator/cheriot_sim /usr/local/bin
+
+      - name: Check sail sim
+        run: cheriot_sim
+
+      - name: Build test runner (host)
+        working-directory: ./cheri/tests/runner
+        run: cargo build --release
+
+      - name: Run internal tests (CHERIoT)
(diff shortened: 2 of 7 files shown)


jacobpake:
@xdoardo I am leaving as a draft as there are a couple of small TODOs, but would appreciate a review particularly on the content of the GHA workflow file, and if you could double check the logs to verify the desired behaviour.

I would like to keep the static configuration files around until I am more confident about the config (after merge), but if you would prefer I can use gen_bootstrap (with or without modification) before merging.

.github/workflows/cheri-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+          opam init --reinit --bypass-check --disable-sandboxing --disable-shell-hook
+          eval $(opam env)
+          opam install sail --confirm-level=unsafe-yes
+          cd cheriot-sail && make csim && sudo cp c_emulator/cheriot_sim /usr/local/bin
+
+      - name: Check sail sim
+        run: cheriot_sim
+
+      - name: Run `x test tests/codgen-llvm` (CHERIoT)
+        run: CC="clang" CXX="clang++" ./x test tests/codegen-llvm --target=riscv32cheriot-unknown-cheriotrtos --skip src/tools/linkchecker --ci=false
```
xdoardo: The `codegen-llvm` tests could probably run before building `sail`

.github/workflows/cheri-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+          path: ~/.cache/ccache
+          key: ccache-${{ runner.arch }}-${{ runner.os }}-${{ github.run_id }}
+          restore-keys: |
+            ccache-${{ runner.arch }}-${{ runner.os }}
+
+      - name: Clear ccache stats
+        run: ccache --zero-stats
+
+      - name: Build LLVM
+        run: CC="clang" CXX="clang++" ./x build llvm --ci=false
```
xdoardo: Are `CC` and `CXX` still necessary?

xdoardo (commented):
> if you could double check the logs to verify the desired behaviour.

I just checked and they seem correct, I think we are doing everything we did in the Cirrus config.

> I would like to keep the static configuration files around until I am more confident about the config (after merge)

Sounds good. I'd like, in the future, to either use our script to generate the bootstrap config in CI too or, alternatively, to remove every mention of CI from `gen_boostrap.sh`. Maybe we can add a "cheri" profile to the default bootstrap configs?

resistor:
> Ccache is used locally and the cache is saved/restored using GitHub Actions Cache. As a side effect, building LLVM is faster.

We need to be a little careful with this, as it will cause a lot of billable ingress/egress traffic from the runners. The original Cirrus-backed version kept the cache local in GCP, which is why I disabled it when moving off of Cirrus.

jacobpake:
> > Ccache is used locally and the cache is saved/restored using GitHub Actions Cache. As a side effect, building LLVM is faster.
> 
> We need to be a little careful with this, as it will cause a lot of billable ingress/egress traffic from the runners. The original Cirrus-backed version kept the cache local in GCP, which is why I disabled it when moving off of Cirrus.

I can disable it if it will be expensive. The last run was ~250MB each direction and our LLVM build was ~90s.

IIRC, WarpBuild can handle actions cache on GCP (I think cirun is AWS only).

I should have added that "building LLVM is faster" in the quote above referred to having the cache on disk vs using the HTTP server with Cirrus.

.github/workflows/cheri-ci.yml
jacobpake: It looks like if we don't set these then it will use `cc`. It can be configured in bootstrap.toml, but I will see if setting these in the top level env causes any problems.

jacobpake: Removed in https://github.com/CHERIoT-Platform/cheri-rust/pull/160/changes/9e07f7b396f60c47657c086911aa31b81727ffc2 and added to top level env in https://github.com/CHERIoT-Platform/cheri-rust/pull/160/changes/82522e5409215661f7bf8e73ccd7f668e9f834d6

Configuring in bootstrap.toml requires knowing the host target, we can handle there in a future version of gen_bootstrap.py

.github/workflows/cheri-ci.yml
jacobpake: Moved in https://github.com/CHERIoT-Platform/cheri-rust/pull/160/commits/8d24eba76cc76adb88c60db3668e643c9b245be3

xdoardo:
(I force-pushed to rebase on top of the commit that bumps LLVM, which was making CI fail)

xdoardo:
@jacobpake Should we merge this or do you think it would be better to wait for comments on https://github.com/CHERIoT-Platform/devcontainer/pull/46?

xdoardo (approved):

.cirun.yml
```diff
@@ -0,0 +1,36 @@
...
+    cloud: gcp
+    instance_type: c4d-standard-32
+    machine_image: ubuntu-2404-noble-amd64-v20260503
+    region: us-east1-b
+    preemptible: true
+    extra_config:
+      baseDisk:
+        diskSizeGb: 120
+    labels:
+      - cirun-gcp-x86
```
xdoardo: If I'm not mistaken we don't use these two anywhere, right? Were they intended to be used to build the artefacts?

.github/workflows/cheri-ci.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Check rust
+        run: cargo --version
+
+      - name: Build Sail simulator
+        run: |
+          git clone --recurse-submodules --depth=1 https://github.com/CHERIoT-Platform/cheriot-sail
+          opam init --reinit --bypass-check --disable-sandboxing --disable-shell-hook
+          eval $(opam env)
+          opam install sail --confirm-level=unsafe-yes
+          cd cheriot-sail && make csim && sudo cp c_emulator/cheriot_sim /usr/local/bin
```
xdoardo, line 84: Just a note for the future: this is pretty slow because it is building _sail_ itself, and then the simulator. I think having aarch64 and x86_64 releases of sail itself would be pretty useful, I'll ask. If we can't convince them, maybe we can try using the releases from https://github.com/rems-project/sail/releases?

jacobpake: I agree and hope they will be convinced. If not, an alternative is to produce it here as an "artifact" and reuse between jobs.

resistor: Why not do the testing inside a devcontainer that already has the simulator in it?

jacobpake:
> @jacobpake Should we merge this or do you think it would be better to wait for comments on [CHERIoT-Platform/devcontainer#46](https://github.com/CHERIoT-Platform/devcontainer/pull/46)?

@xdoardo I think we should merge. I've squashed the commits and left cirrus in, we should open a PR to remove it after we are green on cirun on beta.

.github/workflows/cheri-ci.yml
jacobpake: Good point. How does that work with on GHA with GCP? Does it need to download the full container from gchr (or would we have it in GCP i.e. no ingress cost)? 

Is there any other tradeoff?

.cirun.yml
jacobpake: I removed the two definitions not being used.

.github/workflows/cheri-ci.yml
resistor: I guess I would hope it's cached, but I don't know for certain.

xdoardo: I would suggest to merge this as it is now and polish everything later. It seems that we won't get releases from cheriot-sail, so we'll have to either download the artefact from the sail repository or use a container, but we can think about it after this is merged and we have removed cirrus.

PR #166, opened 2026-05-28
CI: don't build sail, use release

xdoardo:

### DIFF
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           set -eo pipefail
           sudo apt update
-          sudo apt install -y pkg-config clang ninja-build lld cmake perl opam z3 libgmp-dev libssl-dev git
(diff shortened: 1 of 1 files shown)


xdoardo:
Initially using the sail release made it take 25 minutes instead of 35 minutes building sail from scratch. Apparently installing `zlib1g-dev` and/or explicitly setting `CC` and `CXX` reduced it to ~6 minutes. I wonder if we are missing any other dependency that could make this process faster.

jacobpake:
> Initially using the sail release made it take 25 minutes instead of 35 minutes building sail from scratch. Apparently installing `zlib1g-dev` and/or explicitly setting `CC` and `CXX` reduced it to ~6 minutes. I wonder if we are missing any other dependency that could make this process faster.

Probably ccache will help, and we could also keep the Z3 cache between runs (it gets written to file `sail_smt_cache`)

xdoardo:
Right, then I guess it's fine as it is for now: I'll mark this as ready for review.

seharris (approved):

PR #171, opened 2026-06-04
CI experiment (do not merge) (edited title)

seharris:
This is a fake pull request created to try to replicate a CI problem and find workarounds.


seharris:
Conclusions so far: having a PR targeting a branch which is not `beta` (I only tested `old-beta`) and pushing a commit to it gets CI into an odd state where it gets stuck (CI info appears under the CI, but it sits there saying results are expected forever). Changing the target branch to `beta` will then break CI a little more (no CI info appears on the PR). Closing and reopening the PR, or pushing a commit to it looks like it gets CI to work again.

PR #172, opened 2026-06-05
runner: don't print panic message twice in tracing mode

christiankissig:
## Summary

Fixes the regression reported in #38 (reopened): the test runner prints the panic message **twice** when run at trace verbosity.

In tracing verbosity log level, `traceln!` statements output complete `stdout` and `stderr` already. In that case, skip parsing `stdout` for the panic messsage and output.

Updates `traceln!` statements above to display test name, matching the `println!` statements for parsed panic messages below.

### DIFF
--- a/cheri/tests/runner/src/runner/mod.rs
+++ b/cheri/tests/runner/src/runner/mod.rs
@@ -161,16 +161,23 @@ impl App {
 
                 let stderr = String::from_utf8(output.stderr)?;
 
-                traceln!(self, "--- begin stdout from test runner --");
+                traceln!(self, format!("--- begin stdout from test runner ({name}) --"));
                 traceln!(self, stdout);
-                traceln!(self, "--- end stdout from test runner --");
-                traceln!(self, "--- begin stderr from test runner --");
+                traceln!(self, format!("--- end stdout from test runner ({name}) --"));
+                traceln!(self, format!("--- begin stderr from test runner ({name}) --"));
                 traceln!(self, stderr);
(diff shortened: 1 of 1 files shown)


cheri/tests/runner/src/runner/mod.rs
```diff
@@ -161,16 +161,23 @@ impl App {
...
                 traceln!(self, stderr);
-                traceln!(self, "--- end stderr from test runner --");
+                traceln!(self, format!("--- end stderr from test runner ({name}) --"));
+
+                // In trace mode the full stdout (including the panic message) have already been
+                // printed.
+                let in_trace = matches!(
+                    self.verbosity.log_level(),
+                    Some(level) if level >= clap_verbosity_flag::log::Level::Trace
+                );
```
xdoardo:
```suggestion
                let in_trace = self
                    .verbosity
                    .log_level()
                    .is_some_and(|v| v >= clap_verbosity_flag::log::Level::Trace);

```
might be slightly more idiomatic.

christiankissig: I had this first too, but saw there's an open issue on is_some_and https://github.com/rust-lang/rust/issues/143956 for stabilization.

xdoardo: That's for making the function `const`, the function itself has been stable since Rust 1.70.0. Here it probably will never be CTFE'd, since it depends on user parameters. We also already use the same function in the `log` module.

Did you encounter an error while trying it?

xdoardo (approved):

PR #173, opened 2026-06-08
Use `cheri_address_set` instead of doing address arithmetic in `with_addr`

xdoardo:
The [`with_addr`](https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/core/src/ptr/mut_ptr.rs#L203) function in `core::ptr::mut_ptr` replaces the address of a pointer with a new one. Currently, the implementation of this function relies on address arithmetic to compute the diff between the current address and the incoming one, in order to use [`wrapping_byte_offset`](https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/core/src/ptr/mut_ptr.rs#L210) to update the pointer. 

This might be suboptimal in terms of generated code, and on CHERIoT it also resulted in a capability losing its tag in the process (see [this](https://github.com/CHERIoT-Platform/cheri-rust/blob/b9ed1ed19ebb90a197a4bd5d85e82c129a8372fd/library/coretests/tests/ptr.rs#L764) test), due to an LLVM optimisation that changed the semantics (see https://github.com/CHERIoT-Platform/llvm-project/pull/407). The LLVM optimisation is fixed, so in principle it should be ok to leave `with_addr` as it is, but it seems to me that it makes more sense to just lower it to our `cheri_address_set` built-in, which does exactly what the `with_addr` function should do.

### DIFF
--- a/library/core/src/ptr/mut_ptr.rs
+++ b/library/core/src/ptr/mut_ptr.rs
@@ -197,17 +197,30 @@ impl<T: PointeeSized> *mut T {
     /// `self` to the given address, and therefore has all the same capabilities and restrictions.
     ///
     /// This is a [Strict Provenance][crate::ptr#strict-provenance] API.
+    ///
+    /// # Safety
+    ///
+    /// On CHERI platforms just changing the address of a pointer might result in the capability's tag being cleared.
     #[must_use]
     #[inline]
     #[stable(feature = "strict_provenance", since = "1.84.0")]
     pub fn with_addr(self, addr: usize) -> Self {
-        // This should probably be an intrinsic to avoid doing any sort of arithmetic, but
-        // meanwhile, we can implement it with `wrapping_offset`, which preserves the pointer's
-        // provenance.
-        let self_addr = self.addr() as isize;
-        let dest_addr = addr as isize;
-        let offset = dest_addr.wrapping_sub(self_addr);
-        self.wrapping_byte_offset(offset)
+        #[cfg(target_family = "cheri")]
+        // SAFETY: the `cheri_address_set` intrinsic has no prerequisites to be called.
+        unsafe {
+            crate::intrinsics::cheri::cheri_address_set(self, addr) as *mut _
+        }
+
+        #[cfg(not(target_family = "cheri"))]
+        {
+            // This should probably be an intrinsic to avoid doing any sort of arithmetic, but
+            // meanwhile, we can implement it with `wrapping_offset`, which preserves the pointer's
+            // provenance.
+            let self_addr = self.addr() as isize;
+            let dest_addr = addr as isize;
+            let offset = dest_addr.wrapping_sub(self_addr);
+            self.wrapping_byte_offset(offset)
+        }
     }
 
     /// Creates a new pointer by mapping `self`'s address to a new one, preserving the original
--- a/compiler/rustc_hir_analysis/src/check/intrinsic.rs
+++ b/compiler/rustc_hir_analysis/src/check/intrinsic.rs
@@ -819,10 +819,10 @@ pub(crate) fn check_intrinsic_type(
             Ty::new_imm_ptr(tcx, tcx.types.unit),
         ),
         sym::cheri_address_set => (
+            1,
             0,
-            0,
-            vec![Ty::new_imm_ptr(tcx, tcx.types.unit), tcx.types.usize],
-            Ty::new_imm_ptr(tcx, tcx.types.unit),
+            vec![Ty::new_imm_ptr(tcx, param(0)), tcx.types.usize],
+            Ty::new_imm_ptr(tcx, param(0)),
         ),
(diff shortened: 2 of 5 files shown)


library/core/src/ptr/mut_ptr.rs
```diff
@@ -201,13 +201,22 @@ impl<T: PointeeSized> *mut T {
...
     pub fn with_addr(self, addr: usize) -> Self {
-        // This should probably be an intrinsic to avoid doing any sort of arithmetic, but
-        // meanwhile, we can implement it with `wrapping_offset`, which preserves the pointer's
-        // provenance.
-        let self_addr = self.addr() as isize;
-        let dest_addr = addr as isize;
-        let offset = dest_addr.wrapping_sub(self_addr);
-        self.wrapping_byte_offset(offset)
+        #[cfg(target_family = "cheriot")]
+        // SAFETY: If the address is out of the representable bounds of the underlying capability, just changing the address might result in the capability's tag being cleared.
```
christiankissig: NIT `SAFETY:` does not explain why the `unsafe` block is sound, also `SAFETY:` comment should be in `unsafe` block. Did you mean to use `# Safety`  https://std-dev-guide.rust-lang.org/policy/safety-comments.html

library/core/src/ptr/mut_ptr.rs
```diff
@@ -201,13 +201,22 @@ impl<T: PointeeSized> *mut T {
     #[inline]
     #[stable(feature = "strict_provenance", since = "1.84.0")]
     pub fn with_addr(self, addr: usize) -> Self {
```
christiankissig, line 207: This change is only applied to the `mut_ptr`, not `const_ptr` : https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/core/src/ptr/const_ptr.rs#L212 . Doesn't the same tag clearing problem occur for `const_ptr` too?

library/core/src/ptr/mut_ptr.rs
```diff
@@ -201,13 +201,22 @@ impl<T: PointeeSized> *mut T {
...
-        // This should probably be an intrinsic to avoid doing any sort of arithmetic, but
-        // meanwhile, we can implement it with `wrapping_offset`, which preserves the pointer's
-        // provenance.
-        let self_addr = self.addr() as isize;
-        let dest_addr = addr as isize;
-        let offset = dest_addr.wrapping_sub(self_addr);
-        self.wrapping_byte_offset(offset)
+        #[cfg(target_family = "cheriot")]
```
christiankissig: Does the same tag clearing problem and fix apply to the `cheri` target?

library/core/src/ptr/mut_ptr.rs
xdoardo: > Did you mean to use # Safety [std-dev-guide.rust-lang.org/policy/safety-comments.html](https://std-dev-guide.rust-lang.org/policy/safety-comments.html)

There are many `// SAFETY` comments in there, for example:
https://github.com/CHERIoT-Platform/cheri-rust/blob/97f1d5ca8811a246560574cdb7e6ece910622257/library/core/src/ptr/mut_ptr.rs#L476-L479

In fact, tidy will complain about an undocumented unsafe if I use `/// # Safety` or `// # Safety`.

> also SAFETY: comment should be in unsafe block. 

Maybe I'm misunderstanding: are you saying that it should be 
```rust
unsafe {
	/// SAFETY...
}
```
perhaps? It seems that what you linked shows the [opposite](https://std-dev-guide.rust-lang.org/policy/safety-comments.html#:~:text=unsafe%20%7B%20(self.get_unchecked(0..mid)%2C%20self.get_unchecked(mid..self.len()))%20%7D).

> does not explain why the unsafe block is sound.

The reason is as the `arith_offset` above, and I will add that to the comment.

library/core/src/ptr/mut_ptr.rs
xdoardo: I wrongfully assumed that the implementation for `*const T` relied on the implementation for `*mut T`, but that's not the case. Thank you, will fix in the next commit.

library/core/src/ptr/mut_ptr.rs
xdoardo: Ah, I think I see what you mean now: yes, I will add my comment in the doc comment of the function.

library/core/src/ptr/mut_ptr.rs
xdoardo: I was a bit of two minds, as I'd like to be a bit conservative about adding default behaviours, but in this case it seems the right thing to do, changed.

christiankissig (approved):
LGTM

jacobpake:
EDIT: I just re-read the bit about fix in LLVM, disregard

It seems that `ptr::nonnull_tagged_pointer_with_provenance` is passing on current beta. 

If this is the case, do we know why? 

It could be the divergent optimise settings playing tricks again.

PR #182, opened 2026-06-10
Use `address_size()` in `ProvenanceMap` computations

### LINKED ISSUE #181: `ProvenanceMap` uses `pointer_size()` for the size of pointers during CTFE
```shell
$ ./cheri/ci/script/coretest.sh ptr
   Compiling coretests v0.0.0 
error[E0080]: unable to read parts of a pointer from memory at alloc645+0x2
   --> coretests/tests/ptr.rs:961:17
    |
961 |         assert!(*s1.0.ptr == 1);
    |                 ^^^^^^^^^ evaluation of `ptr::test_const_swap_ptr::{constant#0}` failed here
    |
    = help: this code performed an operation that depends on the underlying bytes representing a pointer
    = help: the absolute address of a pointer is not known at compile-time, so such operations are not supported
```

The evaluation failed here:
https://github.com/CHERIoT-Platform/cheri-rust/blob/d8e99edb2f5145184eceb0fc5c88c1fb9417b2bc/compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs#L263-L265

The other code paths that, during CTFE, lead to this insertion use `address_size()` to give the size of what is known at compile-time about the pointer, not `pointer_size()`, making this assertion fail. 

Looks like the simple fix for this would be to just replace `pointer_size()` there (and in the other functions in that module) with `address_size()`. In principle this should work, assuming that what the CTFE machinery does _on_ pointers (not necessarily with the data pointers point to) never shows up at runtime.

xdoardo:
More context in #181.

### DIFF
--- a/compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
+++ b/compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
@@ -83,10 +83,10 @@ impl ProvenanceMap {
 
 impl<Prov: Provenance> ProvenanceMap<Prov> {
     fn adjusted_range_ptrs(range: AllocRange, cx: &impl HasDataLayout) -> Range<Size> {
-        // We have to go back `pointer_size - 1` bytes, as that one would still overlap with
+        // We have to go back `address_size - 1` bytes, as that one would still overlap with
         // the beginning of this range.
         let adjusted_start = Size::from_bytes(
-            range.start.bytes().saturating_sub(cx.data_layout().pointer_size().bytes() - 1),
+            range.start.bytes().saturating_sub(cx.data_layout().address_size().bytes() - 1),
         );
         adjusted_start..range.end()
     }
@@ -136,11 +136,11 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
         range: AllocRange,
         cx: &impl HasDataLayout,
     ) -> impl Iterator<Item = (AllocRange, Prov)> {
-        let ptr_size = cx.data_layout().pointer_size();
+        let addr_size = cx.data_layout().address_size();
         let ptr_provs = self
             .range_ptrs_get(range, cx)
             .iter()
-            .map(move |(offset, p)| (alloc_range(*offset, ptr_size), *p));
+            .map(move |(offset, p)| (alloc_range(*offset, addr_size), *p));
         let byte_provs = self
             .range_bytes_get(range)
             .iter()
@@ -154,13 +154,13 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
         let Some(bytes) = self.bytes.as_deref_mut() else {
             return true;
         };
-        let ptr_size = cx.data_layout().pointer_size();
+        let addr_size = cx.data_layout().address_size();
         while let Some((offset, first_frag)) = bytes.iter().next() {
             let offset = *offset;
             // Check if this fragment starts a pointer.
-            let range = offset..offset + ptr_size;
+            let range = offset..offset + addr_size;
             let frags = bytes.range(range.clone());
-            if frags.len() != ptr_size.bytes_usize() {
+            if frags.len() != addr_size.bytes_usize() {
                 // We can't merge this one, no point in trying to merge the rest.
                 return false;
             }
@@ -189,7 +189,7 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
             return Ok(Some(prov));
         }
         // The other easy case is total absence of provenance, that also always works.
-        let range = alloc_range(offset, cx.data_layout().pointer_size());
+        let range = alloc_range(offset, cx.data_layout().address_size());
         let no_ptrs = self.range_ptrs_is_empty(range, cx);
         if no_ptrs && self.range_bytes_is_empty(range) {
             return Ok(None);
@@ -261,7 +261,7 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
     }
 
     pub fn insert_ptr(&mut self, offset: Size, prov: Prov, cx: &impl HasDataLayout) {
-        debug_assert!(self.range_empty(alloc_range(offset, cx.data_layout().pointer_size()), cx));
+        debug_assert!(self.range_empty(alloc_range(offset, cx.data_layout().address_size()), cx));
         self.ptrs.insert(offset, prov);
     }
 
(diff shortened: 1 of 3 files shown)


compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
```diff
@@ -86,7 +86,7 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
         // We have to go back `pointer_size - 1` bytes, as that one would still overlap with
```
christiankissig: should this say address_size?

seharris (dismissed):
(Approval subject to resolving @christiankissig's comment)

I think this works, but I'd like to check we're on the same page about the implications.

For the Morello port, we took a convservating approach to this problem, and included both data size and overall size in `AllocRange`. That meant that reads and writes of capabilities in const evaluation knew where the address should go, and which bytes should be left empty for metadata. The downside was that it meant touching a lot of code.

For CHERIoT, we're currently trying to avoid touching as much code by not adding new fields to `AllocRange`. This means reads and writes can only reason about where addresses are, not where placeholders for metadata are. I don't think this is necessarily bad, but it does create more opportunities to break things without noticing.

As an example, this is what I think the representation of a struct will be during constant evaluation on CHERIoT, values which can (I think) then be passed on to LLVM and used at run time:
```Rust
struct Example { a: u32, b: u16, b: *const (), c: u64 }
const DUMMY: () = ();
const EXAMPLE: Example = Example { a: 123, b: 456, c: &DUMMY, d: 789 };
// Bytes of `EXAMPLE`:
// 7b 00 00 00 15 03 ## ## ← a, b, padding is uninitialised
// < address > ## ## ## ## ← c, metadata is uninitialised
// c8 01 00 00 00 00 00 00 ← d
```

The metadata part of pointers effectively becomes padding, and is left undefined, which should work. What I don't know, is what happens if you try to contrive a way to write into the undefined bytes. It might be worth testing.

xdoardo:
> What I don't know, is what happens if you try to contrive a way to write into the undefined bytes. It might be worth testing.

Yes, I agree, and that's what I tried to convey in #181. I couldn't come with a meaningful test, and it would make sense to have that before merging this. Do you happen to have any idea what such a test might look like?

seharris:
> Do you happen to have any idea what such a test might look like?

It looks like const evaluation is supposed to reject transmuting pointers in most cases, so if this is possible at all it will probably need to be via a slightly weird edge case. I'm wondering if the rules are more permissive for writes via pointers. I'll do some experiments and see if I can find anything. If I can't, then perhaps can can assume that this is either impossible or obscure enough that it isn't a problem in real use cases?

seharris:
I've run some tests manually and didn't find any sign we've introduced new situations where undefined behaviour can happen without causing errors. We don't have the ui test suite working yet, but when we do we might be able to add the attached sketch as a test to make sure no regressions happen. Let me know if you want me to open an issue to remind us to do that.
[cheri_pointer_misuse.txt](https://github.com/user-attachments/files/29092135/cheri_pointer_misuse.txt)

christiankissig:
> I've run some tests manually and didn't find any sign we've introduced new situations where undefined behaviour can happen without causing errors. We don't have the ui test suite working yet, but when we do we might be able to add the attached sketch as a test to make sure no regressions happen. Let me know if you want me to open an issue to remind us to do that. [cheri_pointer_misuse.txt](https://github.com/user-attachments/files/29092135/cheri_pointer_misuse.txt)

Yes, makes sense to open an issue for that.

compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
christiankissig: @xdoardo  Could you modify the comment to reflect the pointer_ to address_size change with this code change?

compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
```diff
@@ -309,7 +309,7 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
             // No provenance in this range, we are done. This is the common case.
             return;
         }
-        let pointer_size = cx.data_layout().pointer_size();
+        let pointer_size = cx.data_layout().address_size();
```
christiankissig: Should the variable now be named address_size too?

compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
```diff
@@ -385,7 +385,7 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
         cx: &impl HasDataLayout,
     ) -> ProvenanceCopy<Prov> {
         let shift_offset = move |offset| offset - range.start;
-        let ptr_size = cx.data_layout().pointer_size();
+        let ptr_size = cx.data_layout().address_size();
```
christiankissig: Same here: `addr_size`?

compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
```diff
@@ -136,7 +136,7 @@ impl<Prov: Provenance> ProvenanceMap<Prov> {
         range: AllocRange,
         cx: &impl HasDataLayout,
     ) -> impl Iterator<Item = (AllocRange, Prov)> {
-        let ptr_size = cx.data_layout().pointer_size();
+        let ptr_size = cx.data_layout().address_size();
```
christiankissig: Same here: `addr_size`?

compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
christiankissig: Same here: `addr_size`?

seharris:
> Yes, makes sense to open an issue for that.

Done: #204

xdoardo:
Updated to rebase onto `beta` and to address the comments from Christian.

seharris (commented):
Looks good. The only thing I noticed was that I don't think you've addressed Christian's original comment about `// We have to go back 'pointer_size - 1' bytes`.

compiler/rustc_middle/src/mir/interpret/allocation/provenance_map.rs
xdoardo: Sorry, missed it, updated.

christiankissig (approved):
LGTM

seharris (approved):

PR #183, opened 2026-06-11
cheri/runner: pin RTOS commit via CHERIOT_RTOS_COMMIT

### LINKED ISSUE #152: cheri/runner: Pin specific RTOS commit
Sometimes changes to the RTOS may break our runner  for example https://github.com/CHERIoT-Platform/cheriot-rtos/pull/700, which changes the xmake rules for Rust, broke the boilerplate that the runner generates for the tests. In order to make it a bit more stable, we should pin a specific RTOS commit to check out when running the examples.

christiankissig:
Pins commit of `cheriot_rtos` to make build reproducible. Commit ID is taken from the environment variable `CHERIOT_RTOS_COMMIT`, and can be overridden with `--rtos-repo-commit` on `cheri/tests/runner`.

`CHERIOT_RTOS_COMMIT` is applied to `cheri/ci/script/linker.sh`, and `cheri/examples/hello_world/build_and_run.sh`.

The clones are shallow, so a pinned commit generally won't be in the cloned history; each site fetches the commit explicitly (`git fetch --depth=1`) before `git checkout --recurse-submodules`.

An empty value for `CHERIOT_RTOS_COMMIT` signifies an unset value, and the branch tip of `cheriot_rtos` is cloned. Commit is explicitely checked out prior, so that commit selection works with shallow histories.

Closes #152

### DIFF
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -20,6 +20,12 @@ jobs:
     env:
       CC: "clang"
       CXX: "clang++"
+      # Pin the cheriot-rtos tree used by every step that clones it (the test
+      # runner, the coretests linker, the examples) to one commit, so RTOS-side
+      # changes can't silently break CI. Set to a commit SHA to pin; leave empty
+      # to track the branch tip. Override per-invocation with the runner's
+      # `--rtos-repo-commit` flag.
+      CHERIOT_RTOS_COMMIT: ""
     steps:
       - name: Install dependencies
         run: |
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -20,12 +20,7 @@ jobs:
     env:
       CC: "clang"
       CXX: "clang++"
-      # Pin the cheriot-rtos tree used by every step that clones it (the test
-      # runner, the coretests linker, the examples) to one commit, so RTOS-side
-      # changes can't silently break CI. Set to a commit SHA to pin; leave empty
-      # to track the branch tip. Override per-invocation with the runner's
-      # `--rtos-repo-commit` flag.
-      CHERIOT_RTOS_COMMIT: ""
+      CHERIOT_RTOS_COMMIT: "6b3266706786e34a19b6a5c5ac7863f4fc3a588d"
     steps:
       - name: Install dependencies
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/cheri-ci.yml
```diff
@@ -20,12 +20,18 @@ jobs:
     env:
       CC: "clang"
       CXX: "clang++"
+      # Pin the cheriot-rtos tree used by every step that clones it (the test
+      # runner, the coretests linker, the examples) to one commit, so RTOS-side
+      # changes can't silently break CI. Set to a commit SHA to pin; leave empty
+      # to track the branch tip. Override per-invocation with the runner's
+      # `--rtos-repo-commit` flag.
+      CHERIOT_RTOS_COMMIT: ""
```
jacobpake:
```suggestion
      CHERIOT_RTOS_COMMIT: ""
```

This is self-explanatory and the documentation belongs closer to the runner/linker/examples (which you have done).

cheri/tests/runner/src/runner/cmd.rs
jacobpake, line 1: This is quite verbose when taken with the existing git command handling (compare to the xmake equivalent), not a fault of this PR but we should simplify this if possible (either here or in a future PR).

cheri/examples/hello_world/build_and_run.sh
jacobpake, line 1: I tested this command sequence locally and behaved as expected.

jacobpake (changes requested):
I think we could pin a commit in this PR, at the least to confirm the behaviour in CI when the variable is set. If we prefer to keep it empty here, we should do this in another branch.

The latest commit is `a289a4ac15d795960f8f5117f42edbcb7ab9ee0e`.

.github/workflows/cheri-ci.yml
christiankissig: Removed documentation and pinned the latest commit id

cheri/tests/runner/src/runner/cmd.rs
```diff
@@ -103,10 +104,45 @@ impl App {
             }
         };
```
seharris, line 105: This looks like it could be replaced with a call to `self.run_git()`, which would save some duplication. Something like this, I think:

```suggestion
        let branch = branch.map(|v| format!("--branch={v}"));
        let out = format!("{}", out_dir.display());
        let mut args = vec!["clone", url, &out, "--recursive", "--depth=2"];
        if let Some(ref branch) = branch {
            args.push(branch);
        }
        info!(self, "cloning into `", url, "`...");
        self.run_git(&args, &format!("clone `{url}`"))?;
```

xdoardo:
Could you rebase this on top of `beta` and change the target branch?

PR #199, opened 2026-06-17
Pull changes from upstream into `main` (17-06-2026)

xdoardo:


xdoardo:
Looks like GitHub's UI doesn't support the `--rebase-merges` option of `git rebase`, so merging this PR as it is would mean having to use the `merge` mechanism, thus having two merge commits. Since that doesn't look like what we want, we'll use this a way to check that CI is green but directly push this to `main`.

PR #201, opened 2026-06-18
Enable assertions (and more) in CI's bootstrap.toml

xdoardo:
This PR will make CI red. For a bit.

### DIFF
--- a/cheri/bootstrap/ci.toml
+++ b/cheri/bootstrap/ci.toml
@@ -1,12 +1,19 @@
 change-id = "ignore"
 
 [build]
-rustfmt = "./build/host/stage1/bin/rustfmt"
 description = "cheri-rust"
+locked-deps = true
 
 [rust]
 channel = "dev"
+codegen-units-std = 1
+debug-assertions = true
+debug-logging = true
+# See https://github.com/CHERIoT-Platform/cheri-rust/issues/202
+optimize = 3
+overflow-checks = true
 std-features = ["compiler-builtins-mem"]
+verify-llvm-ir = true
 
 [llvm]
 download-ci-llvm = false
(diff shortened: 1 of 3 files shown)


jacobpake:
Whilst we are here, there are some options which upstream uses in CI which we are missing.

These ones seem obvious to include in this PR:

```toml
# Force Cargo to check that Cargo.lock describes the precise dependency
# set that all the Cargo.toml files create, instead of updating it.
build.locked-deps = true

# Whether or not overflow checks are enabled for the compiler and standard
# library.
#
# Defaults to rust.debug value
rust.overflow-checks = true

# Whether to verify generated LLVM IR
rust.verify-llvm-ir = true

# Sets the number of codegen units to build the standard library with,
# regardless of what the codegen-unit setting for the rest of the compiler is.
# NOTE: building with anything other than 1 is known to occasionally have bugs.
rust.codegen-units-std = 1
```

Maybe worth including here:

```toml
# Indicates whether the native libraries linked into Cargo will be statically
# linked or not.
build-cargo-native-static = true
```

Nice to have:

```toml
# Print out how long each bootstrap step took (mostly intended for CI and
# tracking over time)
build.print-step-timings = true
```

And if we want to have some metrics/bookkeeping (not relevant to this PR):

```toml
# Collect information and statistics about the current build, and write it to
# disk. Enabling this has no impact on the resulting build output. The
# schema of the file generated by the build metrics feature is unstable, and
# this is not intended to be used during local development.
build.metrics = true
```

xdoardo:
Following https://github.com/CHERIoT-Platform/cheri-rust/issues/202, I made the `optimize = 3` explicit in all the`bootstrap.toml` files. I also added some of the config flags mentioned above (those I had time to test). Furthermore, I _disabled_ assertions in the install-mode boostrap file: I think debug assertions shouldn't be part of a `rustc-cheri` release, but it can be changed back if we don't agree.

jacobpake (approved):

PR #165, opened 2026-05-28
CI: Autobump LLVM

xdoardo:
This PR introduces an action that runs every ~~Friday at 12pm~~ ("Europe/London" time) to automatically pull changes from our LLVM fork. Example PR: https://github.com/CHERIoT-Platform/cheri-rust/pull/167.

--------
edit: Runs Monday 15:30 UK time.

PR #206, opened 2026-06-23
cheri/ci: add a test runner for coretests and others

jacobpake:
Implements a wrapper/runner around our coretests/libtest to make managing the various submodules and collecting results simpler.

This was previously handled by the bash scripts `cheri/ci/script/coretest.sh` and `cheri/ci/script/coretests.sh`.

This implementation will allow us to easily add new features (such as discussed in #165) as well as unify testing for other suites e.g. alloctests, doctests etc.

The wrapper relies on JSON output from our libtest, so this has been modified to output JSON which is compatible* with the upstream libtest.

This wrapper/runner is not considered complete, but should be at the standard of the scripts it replaces. I suggest that any improvements/enhancements are handled in a future PR.

*common properties are of the same type


cheri/ci/tools/test_runner/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+use clap::Parser;
+
+mod cargo;
+mod results;
+mod runner;
+
+#[derive(Parser)]
+struct Args {
+    /// E.g. "coretests", "alloctests"
+    suite: String,
```
xdoardo, line 37: Would it make sense for `suite` to be an enum? Something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e881a31789bcaf0302949ecf08eec6cd) perhaps?

jacobpake: When there is support for other types of test (e.g. doctests, ui tests), it should definitely need to be an enum. I'm not clear it would be a "subcommand" as in your example, since the arguments should be the same for every suite (currently), but clap lets you have enum arguments as long as they implement from/to string (I was doing this originally!)

I don't think it makes sense right now, as the only thing we do with the string is use it to point Cargo at `library/${suite}/Cargo.toml`, and there should be some separation of concerns between the test runner and the tests we want to use it on (why should it not let you give "alloctests" as an argument?).

xdoardo: Sorry, I copied the wrong URL, [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bb78ab2d3a93f94adaf85ddd46db2e53) is the right one.

> I don't think it makes sense right now, as the only thing we do with the string is use it to point Cargo at library/${suite}/Cargo.toml, and there should be some separation of concerns between the test runner and the tests we want to use it on (why should it not let you give "alloctests" as an argument?).

Right. In the future it might make sense to have the enum type in the `Args` struct, let `clap` do its `FromStr` magic and implement a, say, `to_path` method on the `TestSuite` enum, maybe!

xdoardo:
Ran it locally and looks good to me. Should `coretest.sh` and `coretests.sh` be removed?

cheri/ci/tools/test_runner/src/cargo.rs
```diff
@@ -0,0 +1,205 @@
...
+const TARGET_ABI_CFG: &str = "target.'cfg(target_abi = \"cheriot\")'";
+
+fn linker_config(linker: &str) -> String {
+    format!("{TARGET_ABI_CFG}.linker = '{}'", linker)
+}
+
+const RUSTFLAGS: &str = concat!(
+    "-C embed-bitcode=yes ",
+    "-C panic=abort ",
+    "-C lto ",
```
xdoardo, line 21: Even if it seems not to be a problem (it works in CI and locally as well), I'd suggest to turn LTO off, because it isn't considered to be in a working or safe state right now

cheri/ci/tools/test_runner/src/main.rs
jacobpake: Ah, that example makes sense and is similar to what I did originally. Other suites will need different handling (not just to_path) and I am guessing there will be multiple "categories" of tests so this will probably end up being a little more complicated. We will cross that bridge when we need to add another category of test.

cheri/ci/tools/test_runner/src/cargo.rs
jacobpake: Is the issue on the LLVM side?

Using this flag makes a big difference to the final binary size.

My understanding is that with this flag, the **compiler** will perform optimisations across crates, otherwise it will only optimise within the current crate. 

I don't know how you signal to LLD to perform LTO, but our linker wrapper script strips all arguments / flags (essentially it exists only to extract the list of the object/rlib files we need to hand to xmake).

There is a separate flag which defers the cross-crate optimisations to the linker, which I havent tested: https://doc.rust-lang.org/rustc/codegen-options/index.html#linker-plugin-lto, and I am assuming this is what is unsafe/broken.

If the issue is on the Rust side and/or we are nervous about it then I am happy to remove it, but it is likely it will cause some of the tests to fail because of size.

jacobpake: Turning `lto` on doesn't change the flags sent to `lld`. It's clear that `rustc` performs optimisations before handing off to the linker, but I don't know whether it using any LLVM tools which may be unsafe/broken.

The docs aren't exactly clear:

> This flag controls whether LLVM uses [link time optimizations](https://llvm.org/docs/LinkTimeOptimization.html) to produce better optimized code, using whole-program analysis, at the cost of longer linking time. It takes one of the following values:

Turning it on:

> y, yes, on, true, fat, or no value: perform “fat” LTO which attempts to perform optimizations across all crates within the dependency graph.

And note that even if you omit the flag it _can_ perform lto:

> If -C lto is not specified, then the compiler will attempt to perform “thin local LTO” which performs “thin” LTO on the local crate only across its [codegen units](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units).

The docs for `linker-plugin-lto`:

> The -C linker-plugin-lto flag allows for deferring the LTO optimization to the actual linking step

https://doc.rust-lang.org/rustc/codegen-options/index.html#lto
https://doc.rust-lang.org/rustc/linker-plugin-lto.html

jacobpake:
> Ran it locally and looks good to me. Should `coretest.sh` and `coretests.sh` be removed?

I have removed these, rebased on beta, squashed fixes and also added printing when a test starts (previously only printing when a test finished), which doubles the output but can be very useful for debugging (in the future we can add more/better debugging info and timeouts)

xdoardo (approved):

cheri/ci/tools/test_runner/src/cargo.rs
xdoardo: Thanks for the rundown. Seems that it should be fine as it is!

PR #207, opened 2026-06-23
cheri/ci: separate initial tidy check to run on smaller box

jacobpake:
This PR runs the tidy checks on a smaller machine, before running the full checks on a bigger machine. The benefit of this is that we do not need to spin up a beefy machine for a change which fails on the first check. The downside is that this likely adds 2-5 minutes extra CI time for everything else.


jacobpake:
closing for now but keeping branch as reference in case this is something we want

PR #209, opened 2026-06-23
ci: use sccache

jacobpake:
This PR adds `sccache` to CI, configured to store compiler caches to a Google Cloud Storage bucket. This speeds up building LLVM. The CHERIoT llvm-project are using the same setup.

The cache is currently configured to be shared across all PR events, and separately between pushes to beta.

sccache may also handle Rust compilation units. I do not suspect this will happen without explicitly asking it to do so, but we should check the sccache stats before merging. If this does cache Rust compilations, we need to ensure it is able to distinguish different versions of our stage1 compiler.

### DIFF
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -27,6 +27,25 @@ jobs:
           sudo apt update
           sudo apt install -y pkg-config clang ninja-build lld cmake perl opam z3 libgmp-dev libssl-dev git zlib1g-dev
 
+      - name: Install sccache
+        env:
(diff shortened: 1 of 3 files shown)


jacobpake:
`sccache` is caching some Rust compiles that happen through bootstrap/x.py.

Bootstrap should only use sccache on the stage0 compiler, so this shouldn't be a problem. Upstream is using sccache also.

https://github.com/CHERIoT-Platform/cheri-rust/blob/55e7f63693816ab8562b58f11fa4786d0cf2bb63/src/bootstrap/src/core/build_steps/compile.rs#L1334-L1339

This wont effect anything we do outside of bootstrap/x.py, but worth keeping in mind.

xdoardo (approved):

PR #211, opened 2026-06-24
ci: unify bootstrap configs under profile, reorder workflow

jacobpake:
This PR removes the `ci.toml` and `facade.toml` bootstrap configurations, and replaces them with a bootstrap "profile". The only difference between these two configurations was the inclusion of the "std/optimize_for_size" feature. This PR removes that feature.

The profile can be imported into personal bootstrap configs by adding `profile = "cheriot"` at the top, and that should be the only configuration necessary in order to build for our target. You can add further properties which will override the defaults / profile properties.

This PR makes other minor tweaks to the CI workflow:

 - Removes `x build rustfmt`. Since we are no longer on the beta "channel", we can use a stage0 rustfmt.
 - Moves `x test tidy` ahead of `x build llvm`. As above, we can run this now without building rustfmt. There is no point to building LLVM if we are going to immediately fail on the tidy test.
 - Checks out submodules using upstream's script, which is marginally faster
 - Disables bootstrap/x submodule management as it should be unnecessary (upstream does this)
 - Runs `x setup none` before anything else, this initialises bootstrap and means we don't need to include those logs in any subsequent command


jacobpake:
`x test` fails here, presumably due to using `optimize_for_size`:

https://github.com/CHERIoT-Platform/cheri-rust/actions/runs/28152347261/job/83372732260?pr=211#step:11:425

If this is the case, we may want to decide if we should drop `optimize_for_size` from the `facade` target, to ensure we are testing the same thing we are distributing (we do not use this on our distribution profile currently). This may be easier than adapting the upstream test suite.

jacobpake:
No longer uses `optimize_for_size` to run coretests.

xdoardo (approved):
LGTM. Do you think there is a way to make `./x test` faster?

jacobpake:
> LGTM. Do you think there is a way to make `./x test` faster?

I assume there is not much that can be done from a configuration standpoint except reducing the number of tests or using a bigger/faster machine. When I run it locally it maxes out CPUs for most of the time it is running.

There are a couple of options I have been thinking about:

1. Run it less. E.g. don't run for draft PRs, only run in merge queue. Only run if a PR has a certain label or touches certain files.
2. Cache the results - some suites (e.g. compiletest) will do this when you run them locally. I assume it has a clever way of invalidating the results which could be fine for CI. We can always run them clean at certain stages e.g. in the merge queue.

We might say that we only require the suite to be run in the merge queue, and then have an opt-in (or opt-out) via labels in open PRs. It most cases it is fairly obvious based on the PR whether we expect there to be any meaningful test coverage for the change, and this could be at the discretion of the PR author and/or reviewer.

Examples:

1. I make a change to README.md, I do not run the full suite in my PR. I request a review. The PR is approved and added to the merge queue. The full suite runs.
2. I add a new attribute. I do not run the full suite in my PR, until it is time to request a review. Before requesting a review, I run the full suite. It passes. I request a review.
3. As above but I didn't run the suite. The reviewer suggests running the suite before they approve.

xdoardo:
Makes sense. I guess that's similar to what upstream does with bors. We probably don't want something as complex as that. 

> Only run if a PR has a certain label

This seems the most reasonable choice, and we can change the label depending on the status of the PR as you say.

PR #220, opened 2026-06-24
ci: use local sccache

jacobpake:
Again, but this time using local storage as we don't have an easy and safe way to write to our GCS bucket.

The cache is scoped such that PR branches cannot access each others cache, and `beta` branch cannot access caches from PRs. However, PR branches can access caches created on the `beta` branch, so this will be effective in most cases.

### DIFF
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -37,6 +37,32 @@ jobs:
           # TODO: src/ci/scripts/checkout-submodules.sh should be faster
           submodules: recursive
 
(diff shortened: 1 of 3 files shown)


xdoardo (approved):
The changes look good to me. I don't know enough about `sccache` to spot any pitfalls.

jacobpake:
The major caveat with this approach is that we may end up with loads of useless, large caches in our GitHub Actions Cache, though they will expire eventually. It may be worth limiting the sccache size (this is a sccache setting) in the future.

PR #223, opened 2026-06-25
codegen-llvm tests: fix two codegen tests which match on the offset of fields

### LINKED ISSUE #75: Triage and fix `slice-iter-nonnull.rs`
https://github.com/CHERIoT-Platform/cheri-rust/blob/aedfdf14762fd2ba811bf78a8088127c83736b57/tests/codegen-llvm/slice-iter-nonnull.rs#L20

This test fails because on CHERIoT it generates `i32 8` instead of `i32 4` or `i64 8`. Looking at the definition of the `core::slice::Iter` it seems correct, but this failure could still use a triage.

seharris:
In both cases, the `CHECK` patterns look for a specific offset of a field in either a slice or slice iterator, which ends up differing on CHERIoT because the field is preceeded by a pointer. Differences between 32 and 64 bit targets are already allowed for, but CHERIoT breaks these expectations by having pointers be twice the machine word size.

I've chosen a very simple solution, which keeps size of this change smaller. The downside is that it slightly weakens the test; if a 32 bit target somehow incorrectly calculates the field offset as 8 bytes, it won't be detected. I don't think this is a problem here for these reasons:
* there's only one use of `getelementptr` with `%it` in each test, so this doesn't create an opportunity for another operation to be matched instead and cause a false negative
* I would expect incorrect field offset calculations for slices or structs to be immediately obvious elsewhere (though it isn't obviously covered by another `codegen-llvm` test)
* the offset isn't the focus of either test, so weakening it shouldn't defeat the point of the tests

Let me know if you think I'm being overly optimistic here. The other likely fix would be adding a helper to the tests to get the size of a pointer, and then requiring the offsets to match it, which isn't much messier and would save weakening the pattern matching.

Closes #75.

### DIFF
--- a/tests/codegen-llvm/refs.rs
+++ b/tests/codegen-llvm/refs.rs
@@ -1,5 +1,4 @@
 //@ compile-flags: -C no-prepopulate-passes -Zmir-opt-level=0 -Copt-level=0
-//@ ignore-riscv32cheriot-unknown-cheriotrtos See CHERIoT-Platform/cheri-rust/issues/75
 #![crate_type = "lib"]
 #![no_std]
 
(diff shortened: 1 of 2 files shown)


xdoardo:
>The other likely fix would be adding a helper to the tests to get the size of a pointer, and then requiring the offsets to match it, which isn't much messier and would save weakening the pattern matching

Maybe having something like the `[[ADDRSPACE]]` variable? Say, `[[PTRSIZE_BYTES]]`? If you feel like it would be excessive, let's stash the idea. The patch looks good to me as it is.

seharris:
I thought about doing that and didn't like the idea that we might end up with lots of extra FileCheck variables. If we add one for pointer size in bytes, I can imagine there's a good chance we'd need one in bits, and then possibly some other things, and that feels messy? (given that we'd be accumulating changes to `compiletest`)

xdoardo (approved):

PR #197, opened 2026-06-16
[WIP] Add Github issues from triage of failing tests.

christiankissig:
This adds specific / newly created Github issues to 
* the failing coretests already annotated with `FIXME(cheri/triage)`.

Issue #187, opened 2026-06-15
`coretests`: `test_intersperse_collect_string` "Unhandled" at opt-level 3 (passes at opt-z)


christiankissig:
**Affected test:** `iter::adapters::intersperse::test_intersperse_collect_string` — `intersperse.rs:122` (`#[ignore]`)

**Symptom:** "Unhandled" fault **on opt-3 but not opt-z** - an opt-level-specific
miscompile (an LLVM pass firing at `-O3` introduces the fault).

```rust
let contents = [1, 2, 3];
let s = contents.into_iter().map(|id| id.to_string())
    .intersperse(", ".to_owned()).collect::<String>();
```

The opt-z/opt-3 split points at a single optimisation pass.

**Next step:** diff the `-O3` vs `-Oz` LLVM pass list (`-Cllvm-args=-debug-pass=Arguments`) and bisect. 
Same shape as (`time.rs`, now cleared) - check whether the `time.rs` cleanup commit also fixes this.

jacobpake:
This is passing on opt-3, at least on my alloctest branch, but unnoticed by the runner as the fully qualified name used by `known_issues` has changed at some point.

PR #231, opened 2026-07-01
cheri/testing: run known failures in CI

jacobpake:
Adds a `--known-issues` argument to the test_runner which takes the path a file containing the names of tests which are known to fail, and a link to the relevant GitHub issue.

If a test fails which is included in this list, then it is counted as a fail in the tally, but does not cause the suite as a whole to fail.

A test outcome (pass, ignore, fail) is either expected or unexpected. Failures are always unexpected, unless they are in the known issues, and pass/ignore are always expected, unless they are in the known issues. Only unexpected outcomes are counted as "failures" in CI.

### DIFF
--- a/.github/workflows/cheri-ci.yml
+++ b/.github/workflows/cheri-ci.yml
@@ -112,7 +112,7 @@ jobs:
         run: cargo install --path cheri/ci/tools/test_runner
 
       - name: Run coretests (CHERIoT/facade)
-        run: test-runner coretests
+        run: test-runner coretests --known-issues library/coretests/known_issues
 
       - name: Check disk
         if: always()
--- a/cheri/ci/tools/test_runner/src/known_issues.rs
+++ b/cheri/ci/tools/test_runner/src/known_issues.rs
@@ -1,0 +1,33 @@
+use std::path::Path;
+
+#[derive(Default)]
+pub struct KnownIssues {
+    // Vec is a tad simpler than Set and we don't expect this to be large
+    pub issues: Vec<(String, String)>,
+}
+
+//
+impl KnownIssues {
+    /// The known issues file should have one issue per line in the format `{test_name} {issue}`,
+    /// where `test_name` is the qualified function name printed by the test runner, e.g.
+    /// `num::i32::test_swap_bytes` and `issue` could be a link to an issue on GitHub (or any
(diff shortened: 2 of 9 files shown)


jacobpake:
Example failure when a "known issue" unexpectedly passes: https://github.com/CHERIoT-Platform/cheri-rust/actions/runs/28518729784/job/84536998919?pr=231#step:25:3985

xdoardo:
One thought I had was that starting with https://github.com/CHERIoT-Platform/cheri-rust/pull/197 and continuing with this we're sort of losing the locality of the description of issues, in the sense that it becomes harder and harder to just `grep FIXME\("cheri` and get a hunch of what is the nature of the issues we currently have.  This change might make this slightly harder because now the annotation is not even on the failing test but in another file. 

I don't have a solution for this other than perhaps adding/keeping comments directly on the tests, but it's probably not worth it for such a small QoL improvement.

LGTM!

----
Edit: another thought. Should I create an issue to remember to move the current annotations into the `known_issues` file or are you planning to do that in this PR?

xdoardo (approved):

jacobpake:
> One thought I had was that starting with #197 and continuing with this we're sort of losing the locality of the description of issues, in the sense that it becomes harder and harder to just `grep FIXME\("cheri` and get a hunch of what is the nature of the issues we currently have. This change might make this slightly harder because now the annotation is not even on the failing test but in another file.

I agree, though one could argue that having all of the issues in one place could make this easier.

> I don't have a solution for this other than perhaps adding/keeping comments directly on the tests, but it's probably not worth it for such a small QoL improvement.

I would not be against this, if we don't mind a bit of duplication.

It is possible to add namespaced attributes without modifying the compiler. If we can access those in libtest (without modifying the compiler), perhaps we could have a `cheriot.known_issue = "#187"` attribute instead of this file?

> Edit: another thought. Should I create an issue to remember to move the current annotations into the `known_issues` file or are you planning to do that in this PR?

This PR does this for two tests. We cannot do the same currently for tests which do not compile, and for tests which are slow / hang indefinitely. AFAICT all the remaining annotations are of this type. In the future, we might be able to support running tests with timeouts, but I don't think we can handle tests which fail to compile without some serious work.

Issue #96, opened 2026-02-04
`primitive_docs.rs` doctest assuming usize == ptr size


jacobpake:
https://github.com/CHERIoT-Platform/cheri-rust/blob/6c8a6a78c3a2a0454881a555b7741141039740e6/library/core/src/primitive_docs.rs#L1759

>thread 'main' (1) panicked at library/core/src/primitive_docs.rs:10:1:
assertion `left == right` failed
  left: 8
 right: 4
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
tohost located at 0x8001ee48
Error handler: Unhandled error 0x2 at 0x80010c64 (v:0 0x8000e400-0x8001c600 l:0xe200 o:0x0 p: G R-cgm- X- ---)

PR #232, opened 2026-07-01
core doc tests: skip transmute test that is not compatible

### LINKED ISSUE #95: `transmute` doctest type error
https://github.com/CHERIoT-Platform/cheri-rust/blob/6c8a6a78c3a2a0454881a555b7741141039740e6/library/core/src/intrinsics/mod.rs#L676-L686

Manifesting as a failing doctest (ptr !== usize)

>error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> library/core/src/intrinsics/mod.rs:681:5
    |
681 |     std::mem::transmute::<&i32, usize>(ptr)
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: source type: `&i32` (64 bits)
    = note: target type: `usize` (32 bits)

seharris:
This test demonstrates things to be avoided which super _extra_ don't work on CHERI, so I've just added a note that this is something you shouldn't do on CHERI and marked it to be skipped on CHERI targets.

Closes #95.

### DIFF
--- a/library/core/src/intrinsics/mod.rs
+++ b/library/core/src/intrinsics/mod.rs
@@ -677,9 +677,9 @@ pub const fn forget<T: ?Sized>(_: T);
 /// assert_eq!(num, 0x78563412);
 /// ```
 ///
-/// Turning a pointer into a `usize`:
+/// Turning a pointer into a `usize` (which isn't an option on CHERI targets anyway):
 ///
-/// ```no_run
+/// ```no_run,ignore-cheri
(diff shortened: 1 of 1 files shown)


jacobpake (approved):
Since this "test" is not providing any coverage, and is an example of what *not* to do, I agree that simply ignoring it on our target is the correct thing to do.

jacobpake:
#96 might need similar treatment

PR #237, opened 2026-07-02
rustc_const_eval: use layout type to compute offset when writing scalar pairs

### LINKED ISSUE #236: `rustc_const_eval::write_immediate_to_mplace_no_validate` uses wrong value to calculate offset of `ScalarPair`
https://github.com/CHERIoT-Platform/cheri-rust/blob/3b1d453d5af315ae6eacd92d9f4b926d9d7bac96/compiler/rustc_const_eval/src/interpret/place.rs#L728-L735

In short, when looking at an assignment shaped like `P : (T1, T2) = (a_val, b_val)`, this function uses the `in_memory_size` of `a_val` to figure out where to write `b_val` instead of looking at the `in_memory_size` of `T1`.

This means, for example, that if the compile-time assignment looks like `P: (*const (), usize) = (123usize, 456usize)`, the resulting compile-time value will have 4 bytes of `123usize` followed by 4 bytes of `456usize`, without the 4 bytes of uninitialised padding between them. 

I think this is what actually affects https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/coretests/tests/ptr.rs#L978.

xdoardo:
Should resolve  #236.

### DIFF
--- a/compiler/rustc_const_eval/src/interpret/place.rs
+++ b/compiler/rustc_const_eval/src/interpret/place.rs
@@ -725,14 +725,14 @@ where
                 alloc.write_scalar(alloc_range(Size::ZERO, scalar.data_size()), scalar)?;
             }
             Immediate::ScalarPair(a_val, b_val) => {
-                let BackendRepr::ScalarPair(_a, b) = layout.backend_repr else {
+                let BackendRepr::ScalarPair(a, b) = layout.backend_repr else {
                     span_bug!(
                         self.cur_span(),
                         "write_immediate_to_mplace: invalid ScalarPair layout: {:#?}",
                         layout
                     )
                 };
-                let a_size = a_val.in_memory_size();
+                let a_size = a.in_memory_size(&tcx);
                 let b_offset = a_size.align_to(b.align(&tcx).abi);
                 assert!(b_offset.bytes() > 0); // in `operand_field` we use the offset to tell apart the fields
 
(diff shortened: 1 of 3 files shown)


christiankissig:
This also fixes https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/coretests/tests/slice.rs#L2502

jacobpake:
> This also fixes https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/coretests/tests/slice.rs#L2502

~~@xdoardo if we re-run the checks then CI should pick this up now that known_issues in in beta!~~ nvm, not for a compile error

xdoardo:
> This also fixes [`beta`/library/coretests/tests/slice.rs#L2502](https://github.com/CHERIoT-Platform/cheri-rust/blob/beta/library/coretests/tests/slice.rs#L2502)

Ignore attribute removed with [9fea844](https://github.com/CHERIoT-Platform/cheri-rust/pull/237/commits/9fea8447626e1f50d91ec428b5389bd12007431b).

seharris (approved):
I think this change is reasonable, but I also think there's a chance we'll find this is part of a larger problem.

The size information for `value` comes from a pair of `Scalar` (defined in `rustc_middle`), which can be either an integer value with a size tied to the the size of its numeric value, or a pointer value with a size independent of its value. `layout.backend_repr` works in a similar way. The size information used there comes from `Primitive` (defined in `rustc_abi`), which can be an integer, float, or pointer. I'm a bit worried that if the size we're getting from `a_val` is wrong because `value` for some reason is an integer instead of a pointer, then what is there to prevent the same thing happening to `a`?

I'm not sure there's immediately anything we can or should do about this, but it may be worth keeping it in mind.

PR #244, opened 2026-07-09
cheriot: add compartment/library ccs

jacobpake:
Enables the use of the compartment/library calling conventions for CHERIoT as defined here: https://github.com/CHERIoT-Platform/llvm-project/blob/7267319de4efbca1a8f93d4667f84afe4fd377c2/llvm/include/llvm/IR/CallingConv.h#L290-L298

There are 3 ABIs available, split between 2 feature flags:

`feature(abi_cheriot_library_call)` enables the use of `cheriot-library-call` and `feature(abi_cheriot_compartment_call)` enables the use of `cheriot-compartment-call` and `cheriot-compartment-callee`.

Both are marked as unstable, experimental, internal features. These CCs should not be exposed to end users, or to non-cheriot targets.

I have tested `cheriot-library-call` within a single compartment and this works as expected. We can use this to make some functions from `compiler_builtins` available (see https://github.com/CHERIoT-Platform/cheri-rust/commit/eee6bcf08fecb09428c425e0e31c6089fbc45aa8, not included in this PR).

The cross-compartment CCs are non-standard such that there a separate ABIs for caller/callee, and their use depends on whether the call comes from within a compartment or not. This may be subject to change. Further treatment of these to be left to design work on support for compartments.


xdoardo:
The changes look good, I had some thoughts regarding adding the `RiscvSpec` enum in the `ArchKind` enum in https://github.com/CHERIoT-Platform/cheri-rust/pull/244/changes#diff-6e5e5d43baacc6f500b254078867aa3e62b3c19167046a94ff4ec172355383a1R260-R265 but I think it looks good, and is similar to what is already in there for arm.

Do you think it would be sensible to add codegen tests in, say, `codegen-llvm/cheriot/cheriot_ccs.rs` or something similar?

jacobpake:
> The changes look good, I had some thoughts regarding adding the `RiscvSpec` enum in the `ArchKind` enum in https://github.com/CHERIoT-Platform/cheri-rust/pull/244/changes#diff-6e5e5d43baacc6f500b254078867aa3e62b3c19167046a94ff4ec172355383a1R260-R265 but I think it looks good, and is similar to what is already in there for arm.

I didn't like it, but seems like this is more approrpriate than adding extra guards on the match clauses.

> Do you think it would be sensible to add codegen tests in, say, `codegen-llvm/cheriot/cheriot_ccs.rs` or something similar?

I'm not too familiar with the suite, but I would like to see what the codegen looks like anyway so I will have a go. 

I'm not confident how appropriate it would be, given they are internal features, and we are likely going to need more plumbing, particularly re. the cross compartment calls.

xdoardo:
> I'm not too familiar with the suite, but I would like to see what the codegen looks like anyway so I will have a go.
> 
> I'm not confident how appropriate it would be, given they are internal features, and we are likely going to need more plumbing, particularly re. the cross compartment calls.

I think it's fine to add the tests there, even if they are internal features. The lib CC will, I think, probably be public: I imagine we won't add an ad-hoc attribute for that, but this might change in the future. 

I added a similar test in the mmio PR:  https://github.com/CHERIoT-Platform/cheri-rust/pull/245/changes#diff-6ef14d5e9a00230334f90bd254518b6fe086c5c9f2fd9a62bc21a9bd2e3148b8

compiler/rustc_abi/src/extern_abi.rs
```diff
@@ -326,6 +338,12 @@ impl ExternAbi {
                 // before returning, and hence cannot guarantee a tail call.
                 false
             }
+            Self::CHERIoTCompartmentCall
+            | Self::CHERIoTCompartmentCallee
+            | Self::CHERIoTLibraryCall => {
+                // FIXME(cheri): need to check
```
jacobpake: I'm not clear what the implication is here. Is returning `false` the safe option?

We should be able to tail call library functions (and maybe "guarantee" it?), and perhaps we could tail call same-compartment functions which use these ABIs (though it would be messy to fiddle with that at this stage).

compiler/rustc_ast_passes/src/ast_validation.rs
```diff
@@ -485,6 +485,7 @@ impl<'a> AstValidator<'a> {
                     | CanonAbi::RustTail
                     | CanonAbi::Swift
                     | CanonAbi::Arm(_)
+                    | CanonAbi::CHERIoT(_) // FIXME(cheri): give some thought
```
jacobpake: Is there anything obvious we should be checking here?

compiler/rustc_abi/src/extern_abi.rs
jacobpake: Does "supports_guaranteed_tail_call" mean we can support some attribute like `#[always(tail_call)]`? Does such a thing exist?

If so then I'm not sure we can do that for libcalls, see counter example: https://github.com/CHERIoT-Platform/llvm-project/issues/405

jacobpake: It seems that this is a requirement of MIR checking usage of [become](https://doc.rust-lang.org/stable/std/keyword.become.html).

If this is false it will report unsupported. So seems safe to make this `false` in all cases.

xdoardo: There are other details which aren't captured in this context, such as the callee's interrupt state. It seems that `false` is a sensible starting point.

jacobpake: Yes, and we can revisit if we need to support `become` for a subset of libcalls

compiler/rustc_ast_passes/src/ast_validation.rs
xdoardo: Looking at what's below, I think that in principle it's fine to leave it as it is: for what concerns `async`/`gen` functions, they may come from another Rust shared library/compartment, so they are theoretically legal, and the same can be said for functions marked `safe`. I think that for now it's fine to leave these checks as coarse, we can refine them with time.

xdoardo (approved):
Adding new CCs seems to be mostly a matter of adding new constructors to enums, although there are some questions to be resolved later. It's good that we already have codegen-llvm and runtime tests. LGTM.

xdoardo (approved):

seharris (approved):

PR #245, opened 2026-07-09
Add the `cheriot_mmio` attribute

xdoardo:
This PR can be reviewed, but I plan to add a couple more things before merging it:
 - [x] Error out if the target is not CHERIoT
 - [x] UI tests for permissions coherence checks


compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1626,6 +1628,64 @@ impl<'tcx> CheckAttrVisitor<'tcx> {
...
+        }
+
+        if has_mut && !(has_read && has_cap) {
+            self.dcx().emit_err(diagnostics::CHERIoTCapImportPermissionsCoherence {
+                span,
+                suggestion: &format!("add the missing permissions"),
+                permissions,
+                reason: &format!(
+                    "as they contain  load mut (`{}`) but don't contain both read (`{}`) and cap (`{}`)"
+                    , Perms::MUT_SYM, Perms::READ_SYM, Perms::CAP_SYM
```
seharris: Nitpick:
```suggestion
                    "as they contain load mut (`{}`) but don't contain both read (`{}`) and cap (`{}`)",
                    Perms::MUT_SYM, Perms::READ_SYM, Perms::CAP_SYM
```

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1626,6 +1629,74 @@ impl<'tcx> CheckAttrVisitor<'tcx> {
                 .emit_err(diagnostics::BothOptimizeNoneAndInline { optimize_span, inline_span });
         }
     }
+
+    fn check_cheriot_mmio(&self, _hir_id: HirId, attr: &AttributeKind, attr_span: Span) {
+        if self.tcx.target_spec().llvm_abiname != rustc_target::spec::LlvmAbi::CHERIoT {
+            self.dcx().emit_err(diagnostics::NotCHERIoTTarget {
+                span: attr_span,
+                attr: &format!("cheriot_mmio"),
```
seharris: Is there a reason not to do this, and save allocating memory?
```suggestion
                attr: "cheriot_mmio",
```

tests/codegen-llvm/cheriot/mmio.rs
```diff
@@ -0,0 +1,28 @@
...
+#![no_std]
+#![feature(cheriot_attributes)]
+#![crate_type = "lib"]
+
+#[repr(C)]
+pub struct Uart {
+    field: u32,
+}
+
+unsafe extern "Rust" {
```
seharris, line 14: I think I'd expect a real UART to be `unsafe`, is there a reason for it to be `safe` here?

tests/ui/feature-gates/feature-gate-cheriot-attributes.rs
```diff
@@ -0,0 +1,22 @@
...
+
+
+#[repr(C)]
+pub struct Uart {
+    field: u32,
+}
+
+unsafe extern "Rust" {
+    #[cheriot_mmio(name = "uart", permissions = "R")]     //~ ERROR  the `#[cheriot_mmio]` attribute is an experimental feature
+    pub safe static UART0: Uart;                          //~ ERROR attribute `cheriot_mmio` can be used on CHERIoT targets only
```
seharris: See earlier comment on whether using `safe` makes sense here.

seharris: Should this test be gated to non-CHERIoT targets? I can't see anything that does, and this error won't fire on CHERIoT targets. Ideally we might also want to test that it doesn't, I guess.

seharris (commented):
Looks good, I just have a couple of minor suggestions.

compiler/rustc_passes/src/check_attr.rs
xdoardo: Oops, no, no reason, just a slip-up, will add your suggestion.

tests/codegen-llvm/cheriot/mmio.rs
xdoardo: I wanted to reuse the example we have in the RFC, but doesn't serve a real use here (other than showing that using `safe` doesn't break the expected result of the attribute). I will remove it.

tests/ui/feature-gates/feature-gate-cheriot-attributes.rs
xdoardo: The `attribute...is experimental` is fired even on non-CHERIoT targets; then, when we do the analysis on the attribute itself, we error out if the current target is not CHERIoT. This test wants to check that the feature gate works, and that should work on any target.

jacobpake (approved):
great!

tests/ui/feature-gates/feature-gate-cheriot-attributes.rs
seharris: Right, the `experimental feature` error should fire on any target, but the `CHERIoT targets only` error won't fire on CHERIoT by definition. So if/when we run the UI tests for CHERIoT, the test won't be able to match on the missing second error and will fail.

xdoardo: Since we're not running the ui tests for CHERIoT now, I think it's something we can deal with later. What do you think?

seharris: Yeah, if you'd prefer to do it that way. I don't think it will make a huge amount of difference to anything whether we do it now or later.

seharris (approved):

PR #247, opened 2026-07-20
Pull changes from upstream onto `main` (20-07-26)

xdoardo:


xdoardo:
Adding @jacobpake as a reviewer to take a look at the changes to the allocator interface in `library/std/src/alloc/*`.

jacobpake (approved):
I'm happy with the changes in alloc - as mentioned I think we should make our realloc implementation explicit in the future - even if it is repeating the fallback implementation - for visibility and to ensure nothing changes under our feet.

Next time we should see if we can have conflict resolution / test fix changes as separate commits to the merge commit, so there is less to sift through!!

PR #248, opened 2026-07-20
Updates to README.md

xdoardo:
Some minor changes to the readme. Stopping CI on purpose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,13 @@ or by opening a new discussion in this repository.
 
 ### The state of this project
 
-As of late November 2025, we have a `rustc` that can produce
+As of late July 2026, we have a `rustc` that can produce
 programs for the CHERIoT platform. This `rustc` can then compile the
 `compiler-builtins`, `core` and `alloc` libraries. We are currently in the
-process of testing the functionalities exposed by these libraries.
+process of testing the functionalities exposed by these libraries. Our tests
+mainly involve the `codegen-llvm` and `coretests` test suites. For the latest
+updates and numbers, please refer to the output of the CI job we run for every PR
+and each push to the `main` and `beta` branches.
 
 As of now, CHERIoT-specific bits such as accessing MMIO devices, shared
 objects, defining compartments and using specific calling conventions are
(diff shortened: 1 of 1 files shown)


jacobpake (approved):
I think "LLM contribution" is ambiguous (my loosest and strictest interpretation are quite different), and something we might want/need to make explicit later on, though not a blocking issue and it much clearer than no policy.

README.md
```diff
@@ -71,6 +74,17 @@ As of now, the process of getting a final image for CHERIoT will entail
 
 For an example of all the steps above, take a look at the [cheri/examples/hello_world](./cheri/examples/hello_world) directory. 
 
+
+### Contributions
+Contributions are welcome. If you are planning to add a new CHERI-like target to this project,
+please reach out to us beforehand so the we can plan the best way to do it together.
```
seharris:
```suggestion
please reach out to us beforehand so that we can plan the best way to do it together.
```

README.md
```diff
@@ -71,6 +74,17 @@ As of now, the process of getting a final image for CHERIoT will entail
...
+Otherwise, we will try to find the time to review your PR!
+
+> [!IMPORTANT]
+> Concerning AI contributions: for most CHERIoT things, we have a policy of zero LLM contributions. 
+> In the specific case of Rust, we have to ensure that everything in this repo is something
+> that we can upstream, thus we must also follow the (draft) [upstream policy](https://github.com/rust-lang/rust-forge/pull/1040) on LLM use.
```
seharris: This works fine, but the logic might be a bit clearer if we said something like this:
> This also ensures that we're compliant with the (draft) [upstream policy](https://github.com/rust-lang/rust-forge/pull/1040) on LLM use, which is important to ensure we can upstream our changes in the future.

xdoardo: Updated

README.md
xdoardo: Updated

seharris (approved):

christiankissig (approved):

PR #249, opened 2026-07-20
core doc tests: use pointer type instead of usize to get size of pointer

### LINKED ISSUE #96: `primitive_docs.rs` doctest assuming usize == ptr size

seharris:
This test uses the size of `usize` to check the size of a pointer, so just swap that out for a hopefully obvious choice of pointer sized type.

Closes #96.

### DIFF
--- a/library/core/src/primitive_docs.rs
+++ b/library/core/src/primitive_docs.rs
@@ -1755,7 +1755,7 @@ mod prim_ref {}
 /// assert_eq!(size_of_val(&not_bar_ptr), 0);
 ///
 /// let bar_ptr: fn(i32) = not_bar_ptr; // force coercion to function pointer
-/// assert_eq!(size_of_val(&bar_ptr), size_of::<usize>());
+/// assert_eq!(size_of_val(&bar_ptr), size_of::<&()>());
 ///
 /// let footgun = &bar; // this is a shared reference to the zero-sized type identifying `bar`
 /// ```


xdoardo:
I skimmed through the doc comment and it seems to me that the fact that the reference has the size of a `usize` isn't mentioned anywhere, so this change hopefully doesn't entail changing the semantics as well. Can you confirm that this is the case, just to be sure?

jacobpake:
> I skimmed through the doc comment and it seems to me that the fact that the reference has the size of a `usize` isn't mentioned anywhere, so this change hopefully doesn't entail changing the semantics as well. Can you confirm that this is the case, just to be sure?

When I looked at this previously, I figured a change like this would lose some meaning - previously "the size of a pointer is usize", now "the size of a pointer is the size of a pointer". The former may be incorrect but the latter isn't really saying anything.

seharris:
My reading of the intent here is that it's highlighting that coercing a function value this way yields a normal pointer, rather than saying anything about the size of pointers.

jacobpake:
> My reading of the intent here is that it's highlighting that coercing a function value this way yields a normal pointer, rather than saying anything about the size of pointers.

I prefer this!

xdoardo (approved):

PR #253, opened 2026-07-30
Add the `cheriot_shared_object` attribute

xdoardo:
See https://github.com/CHERIoT-Platform/cheri-rust/discussions/224 for more details.


compiler/rustc_attr_parsing/src/attributes/codegen_attrs.rs
```diff
@@ -910,6 +935,15 @@ impl SingleAttributeParser for CheriotMMIOParser {
...
         }
 
+        let (import_kind, attr_name) = if cx.attr_path.segments.contains(&sym::cheriot_mmio) {
+            (rustc_hir::attrs::CheriotCapImportKind::MMIO, sym::cheriot_mmio.to_string())
+        } else {
+            (
+                rustc_hir::attrs::CheriotCapImportKind::SharedObject,
+                sym::cheriot_shared_object.to_string(),
+            )
+        };
```
seharris: Probably doesn't matter much, but I think doing this might be slightly more robust to future mistakes:
```suggestion
        } else if cx.attr_path.segments.contains(&sym::cheriot_shared_object) {
            (
                rustc_hir::attrs::CheriotCapImportKind::SharedObject,
                sym::cheriot_shared_object.to_string(),
            )
        } else {
            unreachable!()
        };
```

compiler/rustc_codegen_llvm/src/consts.rs
```diff
@@ -418,26 +418,33 @@ impl<'ll> CodegenCx<'ll, '_> {
...
 
-                if let Some(mmio) = self
-                    .tcx
-                    .hir_attrs(hir_id)
-                    .iter()
-                    .find(|v| matches!(v, Attribute::Parsed(AttributeKind::CheriotMMIO(..))))
+                if let Some(cap_import) =
+                    self.tcx.hir_attrs(hir_id).iter().find(|v| {
+                        matches!(v, Attribute::Parsed(AttributeKind::CheriotCapImport(..)))
+                    })
```
seharris, line 424: Do we do anything to handle multiple MMIO or shared object attributes being applied to the same value?
This logic looks like it will silently ignore all but the first attribute.

compiler/rustc_hir/src/attrs/data_structures.rs
```diff
@@ -1058,7 +1067,7 @@ pub enum AttributeKind {
     },
 
     /// Represents `#[cheriot_mmio]`
```
seharris:
```suggestion
    /// Represents `#[cheriot_mmio]` or `#[cheriot_shared_object]`
```

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1657,27 +1666,27 @@ impl<'tcx> CheckAttrVisitor<'tcx> {
...
-                ),
-            });
+                    span,
+                    suggestion: &format!("add the missing permissions"),
+                    permissions,
+                    reason: &format!(
+                        "as they contain load mut (`{}`) but don't contain both read (`{}`) and cap (`{}`)",
+                        Perms::MUT_SYM, Perms::READ_SYM, Perms::CAP_SYM
+                    ),
+                });
```
seharris: This indent is a bit confusing, is it meant to be like this?

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1657,27 +1666,27 @@ impl<'tcx> CheckAttrVisitor<'tcx> {
...
-                ),
-            });
+                    span,
+                    suggestion: &format!("add the missing permissions"),
+                    permissions,
+                    reason: &format!(
+                        "as they contain load global (`{}`) but don't contain both read (`{}`) and cap (`{}`)",
+                        Perms::GLOB_SYM, Perms::READ_SYM, Perms::CAP_SYM
+                    ),
+                });
```
seharris: Odd indent again.

compiler/rustc_passes/src/check_attr.rs
xdoardo: [70c63dc](https://github.com/CHERIoT-Platform/cheri-rust/pull/253/commits/70c63dc8191ee35a2e19212226e1818ddd289b36) curiously `./x fmt` didn't pick it up.

compiler/rustc_passes/src/check_attr.rs
xdoardo: [70c63dc](https://github.com/CHERIoT-Platform/cheri-rust/pull/253/commits/70c63dc8191ee35a2e19212226e1818ddd289b36)

compiler/rustc_codegen_llvm/src/consts.rs
xdoardo: [a0ec217](https://github.com/CHERIoT-Platform/cheri-rust/pull/253/commits/a0ec217b8653b02264b5e91e3a6e82c67ddc4000)

compiler/rustc_passes/src/check_attr.rs
```diff
@@ -1630,24 +1638,43 @@ impl<'tcx> CheckAttrVisitor<'tcx> {
...
+
+        let all_cap_import_attrs = self
+            .tcx
+            .hir_attrs(hir_id)
+            .iter()
+            .filter(|v| matches!(v, Attribute::Parsed(AttributeKind::CheriotCapImport(..))))
+            .collect::<Vec<&Attribute>>();
+
+        if all_cap_import_attrs.len() > 1 {
+            self.dcx().emit_err(diagnostics::MultipleCHERIoTCapImportAttributes { span });
```
seharris: It might be an idea to use the other span here, pointing to only the permissions makes the error a little confusing:
```suggestion
            self.dcx().emit_err(diagnostics::MultipleCHERIoTCapImportAttributes { span: attr_span });
```

xdoardo: Ah, nice catch. I'll have to rewire it a bit, because we lose the span of the original attribute once it's parsed.

xdoardo: Turns out I already stored the full attribute span in the parsed attribute: 
4f90052. `attr_span` in the function actually contained the span of the _item_ the attribute was assigned to.

seharris (approved):

jacobpake (approved):
exciting!!

PR #260, opened 2026-08-13
coretests: add `math` library from RTOS

jacobpake:
Replaces libcall stubs with implementations now in the RTOS `math` library


jacobpake:
TODO: there are some rogue issues/103 comments in tests/num/mod that might be addressed here.

seharris (approved):

PR #273, opened 2026-09-02
facade/test-runner: quality and safety bits around allocations

jacobpake:
This PR addresses a number of issues with the facade/std/test-runner implementation which became apparent whilst running/debugging alloctests. It is effectively the first commit in #256 less the overalignment support.

- stdio no longer uses CString, which IIRC was allocating, which is not great when we are panicking from an allocator error.
- alloc returns a nullptr on failure, which is the encouraged behaviour
- The test-runner C++ wrapper runs some basic allocator checks, and prints the available heap size, before running tests
- Fixes potential re-entrancy problem in the test-runner panic hook
- The test_runner tool will print when it is building / executing a module, which is a tad more useful than no feedback at all when/if it is apparently hanging

### DIFF
--- a/cheri/ci/build/runner.cc
+++ b/cheri/ci/build/runner.cc
@@ -48,23 +48,40 @@ compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval) {
   return ErrorRecoveryBehaviour::ForceUnwind;
 }
 
-extern "C" void cheriot_print(char *str) {
-    for (; *str; ++str) {
-        char c = *str;
-        MMIO_CAPABILITY(Uart, uart)->blocking_write(c);
-	}
+// This doesn't need to be non-blocking, in the simulator `can_write`
+// is always true, and the Rust `stdio` interface which uses this does
+// not require it to be non-blocking, but it is how the interface is designed,
+// so let's pretend.
+extern "C" size_t cheriot_write(uint8_t *buf, size_t len) {
+  auto uart = MMIO_CAPABILITY(Uart, uart);
+  size_t i = 0;
+  while (i < len && uart->can_write()) {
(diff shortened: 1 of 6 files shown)


cheri/ci/build/runner.cc
```diff
@@ -48,23 +48,40 @@ compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval) {
...
+// This doesn't need to be non-blocking, in the simulator `can_write`
+// is always true, and the Rust `stdio` interface which uses this does
+// not require it to be non-blocking, but it is how the interface is designed,
+// so let's pretend.
+extern "C" size_t cheriot_write(uint8_t *buf, size_t len) {
+  auto uart = MMIO_CAPABILITY(Uart, uart);
+  size_t i = 0;
+  while (i < len && uart->can_write()) {
+    uart->data = buf[i++];
+  }
```
seharris, line 60: This is a nitpick, but using `while` rather than `for` here feels a bit odd.

```suggestion
  for (size_t i = 0; i < len && uart->can_write(); i++) {
    uart->data = buf[i];
  }
```

seharris (approved):

cheri/ci/build/runner.cc
jacobpake: Half the reason for this is that we need the `i` in the outer scope. The other half is that, at least for me, `while ... can_write` reads easier than `for ... can_write`.

Is it still odd vs the for loop with the i outside? If it is then please suggest a change like that and I will apply

seharris: Oh, yeah, I didn't pay enough attention to where `i` is used. I think that's fine as is then.

Issue #271, opened 2026-09-01
Open issues with procedure calling convention


jacobpake:
There are two separate issues here, but it makes sense to consider them in parallel:

1. Capabilities in small aggregates are invalidated when passed/returned from function calls.
2. Our calling convention behaviour w.r.t. small aggregates diverges from the spec, and is incompatible with the Clang's behaviour, which in my opinion also diverges from the spec.

By "small aggregate" I mean an aggregate value (e.g. struct) which *can* (not implying *should*) fit in two capability registers.

---

In regards to capabilities becoming invalidated, this occurs when a small aggregate is passed directly using one or two registers. The way which Rust handles this, is to cast such an aggregate to/from XLEN-sized integers, which invalidates any capability. 

This can be resolved either by passing any such aggregate indirectly (which would break in FFI and is non-optimal), or we can address this by changing the casting behaviour for pointers on our target. The simplest implementation adds a capabality `CastTarget`/`RegKind`, which is how this is handled for targets with floating-point registers.

---

In regards to the divergant ABIs, our current behaviour follows the RISC-V spec, but treats XLEN as 64, such that an `i64` is passed using a single argument register, and an `{i64, i64}` is passed directly using two registers.

The current behaviour in Clang is to follow the RISC-V spec, treating XLEN as 32 bits, except for small aggregates which contain a pointer, in which case 64-bit non-capability data contained in the same aggregate is passed directly. For example, an `{i64, i64}` is passed indirectly, but a `[i64, ptr]` is passed directly (I think the inconsistency here is a bug). The desired behaviour is that `[ptr]` and `[ptr, ptr]` can be passed through registers, as can `[i32]` and `[i32, i32]`.

To resolve this, we will need to clarify the desired behaviour w.r.t. small aggregates, treat XLEN as 32 bits, and introduce special/separate handling for pointer types.

---

In regards to the specification, it is not clear to me how small aggregates containing capabilities and non-capabilities (e.g. `[ptr, i32]`) should be treated. My opinion is that we should specify something similar to the FP convention here (an `[f32, i32]` uses both a floating-point and an integer register).

[...]

PR #281, opened 2026-09-17
Check that thread local storage is not used on CHERIoT

seharris:
`std` provides the macro `thread_local!()` which makes sure a static value is thread local. On CHERIoT, we currently use a stub implementation that assumes we don't have threads. We actually do have threads, so as and when we provide more support for them this may lead to strange behaviour during testing. This change adds a slightly hacky check to the stub implementation, which uses the RTOS thread ID to ensure that the contained values are only ever used from one thread.

I'm not entirely convinced this is worth the bother at the moment, given we're not thinking of doing anything with threads anytime soon. At the same time, I dislike the idea of having a lurking opportunity for weird bugs to materialise in the future. Thoughts?


jacobpake:
I suspected that the error in CI was related to #271 and the changes to EagerStorage, but looks like it is reproducable even with:

```rs
unsafe extern "C" {
    fn thread_id_get_wrapper() -> u16;
}

fn main() {
    let id = unsafe { thread_id_get_wrapper() };
    println!("{}", id);
}
```

But it's fine to pass as a u32:

```c
extern "C" size_t thread_id_get_wrapper(void) {
  return (size_t)thread_id_get();
}
```

```rs
unsafe extern "C" {
    fn thread_id_get_wrapper() -> u32;
}
```

jacobpake:
Oh, it's a libcall, and maybe the wrapper is breaking the CC.

This works without the wrapper:

```rs
#![feature(abi_cheriot_library_call)]

unsafe extern "cheriot-library-call" {
    #[link_name = "_Z13thread_id_getv"]
    fn thread_id_get() -> u16;
}

fn main() {
    let id = unsafe { thread_id_get() };
    println!("{}", id);
}
```

PR #282, opened 2026-09-18
Add the `#[cheriot_compartment]` attribute

xdoardo:
As per title. One thing to keep in mind: there is a change to the RTOS that is needed for this attribute to work correctly which I added as a patch in the new example. I will raise a PR to the RTOS before we merge this to figure out what's the correct solution: don't merge this before we sort out the RTOS side of this feature!

In the meantime, the other changes should be good to review.


cheri/examples/cheri_compartment_attr/crates/cheriot/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(not(target_family = "cheriot"))]
+compile_error!("This crate can only compile to CHERIoT!");
+
+extern crate alloc;
+
+mod cheriot_alloc;
+mod print;
+
+#[allow(unused_imports)]
```
seharris, line 18: I don't think this should be needed for public exports in a library, and doesn't seem to cause issues if I delete it.

cheri/examples/cheri_compartment_attr/crates/cheriot/src/cheriot_alloc.rs
```diff
@@ -0,0 +1,24 @@
+//! Our allocator.
+
+extern crate alloc;
```
seharris, line 3: This doesn't need to be copied into every module when it's in `lib.rs`: <https://doc.rust-lang.org/reference/items/extern-crates.html#r-items.extern-crate.extern-prelude>

cheri/examples/cheri_compartment_attr/crates/cheriot/src/print.rs
```diff
@@ -0,0 +1,34 @@
+//! Our reproduction of the `print!` and `println!` macros.
+
+extern crate alloc;
```
seharris, line 3: This shouldn't be needed (see comment on `cheriot_alloc.rs`).

cheri/examples/cheri_compartment_attr/crates/cheriot/src/print.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    unsafe {
+        cheriot_print_str(str.as_ptr());
+    }
+
+    drop(str);
+}
+
+#[allow(unused)]
+pub use {print, println};
```
seharris, line 34: This is meant to allow the macros to be imported, right? I don't think this is actually needed here, because it only affects imports inside the same crate, and the print macros aren't being used anywhere. Everything still seems to compile if I delete this. If this *is* needed, it might be worth adding a comment, because this an area where Rust gets a bit weird and I think this might confuse people.

cheri/examples/cheri_compartment_attr/crates/my_app/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#![no_std]
+#![feature(cheriot_attributes)]
+#![cheri_compartment = "my_app"]
+
+#[cfg(not(target_family = "cheriot"))]
+compile_error!("This crate can only compile to CHERIoT!");
+
+#[allow(unused)]
+use cheriot::*;
```
seharris, line 12: I'm guessing this is here to demonstrate/test that you can have a compartment import code from multiple other compartments? If that's true, I'd suggest adding a comment to make it clearer, because at the moment this unused import seems a bit weird.

cheri/examples/cheri_compartment_attr/rtos_shim.cc
```diff
@@ -0,0 +1,30 @@
...
+extern "C" void cheriot_print_str(char *s)
+{
+	printf("%s", s);
+}
+
+extern "C" void *cheriot_alloc(size_t size)
+{
+	// debug_log("Trying to allocate {} bytes!", size);
+	Timeout timeout{5};
+	void   *ret = heap_allocate(&timeout, MALLOC_CAPABILITY, size);
```
seharris, line 18: This should probably use the Jacob's updated version which handles errors more robustly:
```suggestion
    // wait for revocation, don't wait for free
    void *ret = heap_allocate(TimeoutWaitForever, MALLOC_CAPABILITY, size,
        AllocateWaitRevocationNeeded);
```

cheri/examples/cheri_compartment_attr/xmake.lua
seharris, line 1: If this example is intended for end users, would it make sense to add some more explanation of what all the options are?

Without going and digging around in the RTOS code and documentation, there are a couple of things here which look potentially important, but without it being obvious to me what they mean:
* `override = true` and `force = true` in `add_files()`
* `{expand = false}` in `on_load()`

compiler/rustc_attr_parsing/src/diagnostics.rs
```diff
@@ -2064,3 +2064,10 @@ pub(crate) struct UnusedDuplicate {
     )]
     pub warning: bool,
 }
+
+#[derive(Diagnostic)]
+#[diag("the compartment name in `#[cheri_compartment = \"<name>\"] cannot be empty")]
```
seharris, line 2069: This is probably fine as is, but we could possibly make it more consistent with existing errors:
```suggestion
#[diag("the compartment name in `#[cheri_compartment = \"...\"] cannot be empty")]
```

compiler/rustc_codegen_ssa/src/codegen_attrs.rs
```diff
@@ -413,6 +416,23 @@ fn apply_overrides(tcx: TyCtxt<'_>, did: LocalDefId, codegen_fn_attrs: &mut Code
             codegen_fn_attrs.flags |= CodegenFnAttrFlags::NO_MANGLE;
         }
     }
+
+    // CHERI-specific.
```
seharris, line 420: I think it'd be good to include a bit more explanation of what this logic is designed to do, and match the comment-before-block pattern other logic in this function is using. Something like:
```suggestion
    // CHERI-specific.
    // `extern` functions with an explicit CHERIoT compartment use that
    // compartment, otherwise they inherit the compartment used by the
    /// containing `extern` block or crate (if any).
```

compiler/rustc_attr_ir/src/data_structures.rs
```diff
@@ -1058,7 +1058,14 @@ pub enum AttributeKind {
         encoding: Symbol,
     },
 
-    /// Represents `#[cheriot_mmio]` or `#[cheriot_shared_object]`
+    /// Represents `#[cheri_compartment]`.
+    CheriCompartment {
+        compartment_name: Symbol,
+        compartment_name_span: Span,
```
seharris, line 1064: This doesn't seem to be read anywhere.

compiler/rustc_ty_utils/src/abi.rs
```diff
@@ -549,7 +549,53 @@ fn fn_abi_new_uncached<'tcx>(
...
+                (Some(fn_c), Some(k_c)) => {
+                    if fn_c == k_c {
+                        conv =
+                            rustc_abi::CanonAbi::CHERIoT(rustc_abi::CHERIoTCall::CompartmentCallee);
+                    } else {
+                        conv =
+                            rustc_abi::CanonAbi::CHERIoT(rustc_abi::CHERIoTCall::CompartmentCall);
+                    }
+                }
+                (None, None) => {}
```
seharris, line 595: This looks reasonable, but I don't feel like I have a good enough handle on how the calling conventions are meant to work to be completely sure. I might take a look at what Clang does to double check we're doing the right things.

tests/codegen-llvm/cheriot/compartment-attribute.rs
```diff
@@ -0,0 +1,88 @@
...
+        // CHECK: call cheriot_compartmentcalleecc addrspace(200) void @other_comp1_entrypoint()
+        other_comp1_entrypoint();
+    }
+
+    // Public within the `prv` module but not exposed externally; normal CC.
+    // CHECK: call addrspace(200) void @_RNvNtCsggRB0ItIQTK_21compartment_attribute3prv17not_an_entrypoint()
+    prv::not_an_entrypoint();
+}
+
+mod prv {
```
seharris, line 49: Would it be an idea to also do this for public visibility? I'm not really sure if this ought to work, name mangling might make it impossible to use this, I guess?
```suggestion
pub mod public_module {
    // Public, just via a slightly less direct route.
    #[no_mangle]
    // CHECK: define dso_local cheriot_compartmentcalleecc void @nested_entrypoint() unnamed_addr addrspace(200) #0
    pub fn nested_entrypoint() {}
}

mod prv {
```

tests/ui/cheri/cheriot/compartment_attribute.rs
seharris, line 1: We could also test the empty compartment name error here:
```Rust
// We don't allow an empty compartment name.
#[cheri_compartment = ""] //~ ERROR the compartment name in `#[cheri_compartment = \"<name>\"] cannot be empty
unsafe extern "C" fn missing_compartment_name();
```

seharris (commented):
This ended up being a lot comments, but I think they're all fairly minor.

